PR #197, opened 2021-06-11
Made some fixes for CASA 6.

1054:
[DO NOT MERGE, STILL DOING TESTS]

Adapting the pipeline to CASA 6 (Python 3). A few changes are made, but tests are running, so please do not merge for now. Comments and code reviews are welcome!

PR #199, opened 2021-07-07
Single-dish Handler

1054:
**_[ON-GOING PROJECT, DO NOT MERGE]_**

Single-dish handler is now added to our pipeline. It can process ALMA total power raw data and obtain image cubes, as long as the total power raw data are defined in the `"ms_file_key.txt"` and the target is defined in the `"singledish_key.txt"`. The latter defines a single-dish FITS file to be feathered with the interferometric data. The single-dish handler will check if the single-dish FITS file is found there or not. If not, the code will start to process the raw data as defined in the `"ms_file_key.txt"` under a working directory `singledish_root+os.sep+target+os.sep+'processing_singledish_'+target`, then copy the final image cube to the desired place. If the single-dish FITS file already exists, the single-dish handler will just say that the file is found and will not be re-processed. 

Preliminary tests with CASA 4 and CASA 5 both now worked, producing very similar image cubes. 

But there are still tests to do and details to discuss. For example, 
- adjusting the working directory path if needed, 
- dosplitants in CASA 5, 
- cleaning up intermediate files, 
- galaxy-specified flagging files, 
- telluric ozone line issue?
- what if multiple tp raw data exist?

Below are some example contents to add to the key files so as to make the single-dish handler working:

Example `"ms_file_key.txt"`
```
ngc1808  2017.1.00984.S  all  12m  1  2017.1.00984.S/science_goal.uid___A001_X1296_Xa50/group.uid___A001_X1296_Xa51/member.uid___A001_X1296_Xa52/calibrated/uid___A002_Xccea8d_X838d.ms
ngc1808  2017.1.00984.S  all  7m  1  2017.1.00984.S/science_goal.uid___A001_X1296_Xa50/group.uid___A001_X1296_Xa51/member.uid___A001_X1296_Xa54/calibrated/uid___A002_Xc64412_X100f.ms
ngc1808  2017.1.00984.S  all  tp  1  2017.1.00984.S/science_goal.uid___A001_X1296_Xa50/group.uid___A001_X1296_Xa51/member.uid___A001_X1296_Xa54/calibrated/uid___A002_Xc64412_X100f.ms

```

Example line to add into `"config_definitions.txt"`
```
singledish_config  tp  {'bl_order':1,'chan_dv_kms':1.0}
```

And below example commands to run the single-dish processing in CASA:
```
import os, sys
sys.path.insert(1, '/data/beegfs/astro-storage/groups/schinnerer/dzliu/PHANGS/Data/SingleDishHandler/scripts/phangs_imaging_scripts_3/analysis_scripts')
sys.path.insert(1, '/data/beegfs/astro-storage/groups/schinnerer/dzliu/PHANGS/Data/SingleDishHandler/scripts/phangs_imaging_scripts_3/phangsPipeline')
sys.path.insert(1, '/data/beegfs/astro-storage/groups/schinnerer/dzliu/PHANGS/Data/SingleDishHandler/scripts/phangs_imaging_scripts_3')

key_file = '/data/beegfs/astro-storage/groups/schinnerer/dzliu/PHANGS/Data/SingleDishHandler/keys/master_key.txt'

from phangsPipeline import phangsLogger as pl
pl.setup_logger(level='DEBUG', logfile=None)

from phangsPipeline import handlerKeys as kh
from phangsPipeline import handlerSingleDish as sdh

this_kh = kh.KeyHandler(master_key=key_file)
this_sdh = sdh.SingleDishHandler(key_handler=this_kh)

this_sdh.loop_singledish()
```

### DIFF
--- a/phangsPipeline/casaImagingRoutines.py
+++ b/phangsPipeline/casaImagingRoutines.py
@@ -687,6 +687,18 @@ def clean_loop(
         copy_imaging(
             input_root=working_call.get_param('imagename'),
             output_root=working_call.get_param('imagename')+'_prev')
+    
+        # Check user-preset mask parameter, disable it if a *.mask already exists
+        
+        if (loop > 0) and \
+           (working_call.get_param('usemask') == "user") and \
+           (working_call.get_param('mask') is not None) and \
+           (working_call.get_param('mask') != ''):
+            if os.path.isdir(working_call.get_param('imagename')+'.mask'+suffix):
+                logger.debug("Found clean mask \"%s\", will not re-use the mask \"%s\" in the clean parameter file."%(\
+                    working_call.get_param('imagename')+'.mask'+suffix, 
+                    working_call.get_param('mask')))
+                working_call.set_param('mask', '')
 
         # Execute the clean call.
 
--- a/phangsPipeline/casaImagingRoutines.py
+++ b/phangsPipeline/casaImagingRoutines.py
@@ -66,7 +66,7 @@ def estimate_cell_and_imsize(
         for kk in range(3):
             for jj in range(3):
                 valid_sizes.append(2 ** (ii+1) * 5 ** jj * 3 ** kk)
-    valid_sizes.sort()
+    valid_sizes = sorted(valid_sizes)
     valid_sizes = np.array(valid_sizes)
 
     # Cell size implied by baseline distribution from analysis
--- a/phangsPipeline/casaMaskingRoutines.py
+++ b/phangsPipeline/casaMaskingRoutines.py
@@ -379,8 +379,8 @@ def signal_mask(
         mask = (mask + old_mask) > 0
     if operation == 'NEW':
         mask = mask
-
-    del old_mask
+    else:
+        del old_mask
 
     logger.info('Recasting as an int.')
     # this might be better: mask.astype(np.int, copy=False)
--- a/phangsPipeline/casaSingleDishRoutines.py
+++ b/phangsPipeline/casaSingleDishRoutines.py
@@ -164,15 +164,15 @@ def getDataColumnForSDBaseline(vis):
     """
     mytb = au.createCasaTool(casaStuff.tbtool)
     mytb.open(vis)
-    names = mytb.colnames()
+    names = copy.copy(mytb.colnames())
     mytb.close()
     columns = []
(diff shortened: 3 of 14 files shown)


1054:
Now trying to include the fix for CASA 6 (pull #197) also in this pull, so that later we only need to merge this pull. 

Some changes:

- `.*.sort()` --> `sorted(.*)` in many scripts including "clean_call.py"
- `freqtol = ''` --> `freqtol = '0.5MHz'` in the definition of `batch_extract_line` in "casaVisRoutines.py", otherwise sometimes extracted line spw from different projects have a ~0.2 MHz shift. 
- some overwriting stuff in "handlerVis.py"
- disabled the `casa_enabled` tag, just use `dry_run`
- fixed an issue where `del old_mask` in "casaMaskingRoutines.py" causes error when `do_multiscale_clean = False` in "handlerImaging.py" (thanks Ivana for testing)

1054:
The `freqtol` change has just been cancelled in the latest commit  :)

thomaswilliamsastro:
Just managed to get this running for the first time, I'll probably come back and edit things as I find them. First impressions:

- It wasn't clear the `ms_file_key.txt` should point to the untarred, raw directory. I think it might make more philosophical sense to have a `singledish_dir_key.txt` or something to help provide some distinction between how this works and the interferometric MSs
- Currently, the `singledish_config` isn't read in (and throws up an error). The `bl_order` is forced to 1 and `chan_dv_kms` comes from the line product definition (I think this one makes sense, but `bl_order` should be exposed properly)
- Possible makes sense to also expose `line_wing` from the `task_execute_single_dish_pipeline`? It seems like something that may need to be tuned per-user
- For multiple TP data for a single target (i.e. different lines), I totally agree on specifying the line product instead of the observation number in `ms_file_key.txt`. I think the fact that you need to point this at a directory and a line instead of observations to concat also makes sense to break out to a `singledish_dir_key.txt`

I'll come back and add things as I find them, but I have the data calibrating now!

thomaswilliamsastro:
Additional issues (not blockers):

1. For MSs with multiple targets, it'll only pull out the first one. This seems like a very fringe case but here for posterity
2. If the velocity resolution is smaller than the native resolution the pipeline crashes. Should probably do a check and set to the minimum if the requested resolution<data resolution

PR #204, opened 2021-09-15
avoid syntax error on load

low-sky:
HI @astrojysun -- I reverted a change here that was causing syntax errors on CASA 5.6.1 with the relative imports.  This seems to be related to how we initialize path in the run files (i.e., which I used sys.path.append), but I'm unclear on what's causing it.  Is there a better way to approach this in the run files or a more global fix (e.g., with the `__init__.py` file?

cc: @tonywong94

### DIFF
--- a/phangsPipeline/utilsLines.py
+++ b/phangsPipeline/utilsLines.py
@@ -5,7 +5,7 @@ import logging
 
 import numpy as np
 
-import .utilsLists as lists
+import phangsPipeline.utilsLists as lists
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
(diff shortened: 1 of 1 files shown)


phangsPipeline/utilsLines.py
```diff
@@ -5,7 +5,7 @@
 
 import numpy as np
 
-import .utilsLists as lists
+import phangsPipeline.utilsLists as lists
```
astrojysun, line 8: `from . import utilsLists as lists`

astrojysun (commented):
This is indeed a typo... I have suggested a corrected line to do the relative import. Thank you for catching this and pinning me!

PR #207, opened 2021-10-19
Add in py3 compatible packaging

e-koch:
Updates the packaging infrastructure to use astropy's cookie-cutter template so the pipeline will eventually by packageable.


e-koch:
The pipeline now runs in casa 6.4! I have finished a full imaging test locally using some HI VLA data in the casa 6.4 (monolithic) and everything ran without obvious issues. The pip install also works fine (but see below).

But the package will not yet work with a pip install (so in a normal python environment) because of analysisUtils, which won't recognize a non monolithic environment as "valid" (from what I can tell). I think we should move ahead with merging most in since this update includes most of the changes needed to migrate to new casa versions, and should be relatively future proof at this point.

When running in a normal python environment becomes possible, **the package will only work with casa >6.4 due to it supporting python 3.8**. The astropy/numpy/scipy environment has dropped python 3.6 support, and that includes critical updates and fixes in the spectral-cube package.

In moving to python 3, some things need to be dropped:
* `reload` is no longer supported. I have commented out those lines for now.
* dictionary keys are not natively a list in python 3, and need to be wrapped in `list(dict.keys())`

@akleroy @low-sky -- someone should have a quick look through the changes (most are whitespace removals... sorry) to check for typos. Otherwise, this is ready to merge!

PR #210, opened 2021-12-23
Allow more flexible continuum extraction & add support for Hydrogen RRLs

astrojysun:
The first two commits added Hydrogen RRLs (as a line family) to the supported line list. Only those located in the ALMA bandpasses were added. Also, they fixed a bug and cleaned up the code format.


astrojysun:
The five new commits allow one to specify frequency ranges and channel width for continuum extraction. They also fixed a couple bugs and addressed code formatting issues in `handlerVis` and `casaVisRoutines`.

PR #213, opened 2022-01-13
Bugfix with nchan/binfactor mismatch to make sure it always divides nicely

thomaswilliamsastro:
Sometimes nchan doesn't divide nicely by binfactor in the image staging steps, which can cause problems where final channels in the cube can be missing data. This should fix that


thomaswilliamsastro:
This is merged in #214

PR #217, opened 2022-02-25
ALMA Download Handler

thomaswilliamsastro:
Add ALMA archive query/download to the pipeline. This is intended to automate archive hunting/data calibration, which is now the most hands-on bit of the whole ALMA data reduction using the pipeline

### DIFF
--- a/phangsPipeline/__init__.py
+++ b/phangsPipeline/__init__.py
@@ -9,10 +9,11 @@ from ._astropy_init import *   # noqa
 from .phangsLogger import setup_logger
 from .handlerKeys import KeyHandler
 from .handlerSingleDish import SingleDishHandler
+from .handlerAlmaDownload import AlmaDownloadHandler
 from .handlerVis import VisHandler
 from .handlerImaging import ImagingHandler
 from .handlerPostprocess import PostProcessHandler
 from .handlerDerived import DerivedHandler
 
-__all__ = ["setup_logger", "KeyHandler", "SingleDishHandler", "VisHandler", "ImagingHandler", "PostProcessHandler",
-           "DerivedHandler"]
+__all__ = ["setup_logger", "KeyHandler", "SingleDishHandler", "AlmaDownloadHandler", "VisHandler", "ImagingHandler",
+           "PostProcessHandler", "DerivedHandler"]
--- a/phangsPipeline/handlerAlmaDownload.py
+++ b/phangsPipeline/handlerAlmaDownload.py
@@ -730,6 +730,10 @@ if has_imports:
             ms_root = self._kh._ms_roots[0]
             dl_dir = os.path.join(ms_root, target, config)
 
+            if not os.path.exists(dl_dir):
+                logger.warning('Directory for %s, %s does not exist. Returning' % (target, config))
+                return None
+
             original_dir = os.getcwd()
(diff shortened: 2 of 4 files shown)


thomaswilliamsastro:
The idea for this is:

- Run through targets, download data based on velocity resolution/requested line/other supplied restrictions
- Pull out the CASA pipeline version and run scriptForPI
- Make all the various key files (e.g. target, linmos, ms_key, sd_key)

It will also download TP data and sort out keys in such a way as to pass off to the singledish handler.

Run like:

```
key_handler = ppl.KeyHandler(master_key=master_key)
dl_handler = ppl.AlmaDownloadHandler(key_handler=key_handler, master_key=master_key)

dl_handler.loop_alma_download(do_all=True,
                                  # allow_proprietary=True,
                                  # username='tgwilliams',
                                  )
# After doing this, restart up the key handler
key_handler = ppl.KeyHandler(master_key=master_key)
```

Where you can flip `allow_proprietary` and `username` to download non-public data. To handle casa versions, there's now a `casaversion_key`, which looks like:

```
# Column 1: CASA version (e.g. 5.4.0-70)
# Column 2: Path to this CASA version (e.g. /usr/local2/misc/casa/casa-release-5.4.0-70.el7/bin)

4.2.2       /data/beegfs/astro-storage/groups/schinnerer/williams/casa/casapy-42.2.30986-pipe-1-64b
4.7.2       /usr/local2/misc/casa/casa-release-4.7.2-el7/bin
5.4.0-70    /usr/local2/misc/casa/casa-release-5.4.0-70.el7/bin
5.6.1-8     /usr/local2/misc/casa/casa-pipeline-release-5.6.1-8.el7/bin
5.7.2-4     /data/beegfs/astro-storage/groups/schinnerer/williams/casa/casa-release-5.7.2-4.el7/bin
```

and for more granular controls over what to download, a `alma_download_key`, e.g.:

```
# Parameters are target ('all' is OK), product ('all' is OK),
# config ('all') is OK, and then restrictions as a dictionary.

ngc0253     all     all     {'proposal_id':'2018.1.01321.S'}
ngc0253     all     12m     {'proposal_id':'2017.1.00161.L'}
ngc0628     all     all     {'proposal_id':'2012.1.00650.S'}
ngc4826     all     all     {'proposal_id':'2017.1.00886.L'}
```

In terms of grouping mosaics, there are 3 options (set by `split_ms`): 'mosaic', 'join', and 'separate'. 'mosaic' is the default and will group observations by project ID/science goal, so should be good for nice uniform data. 'join' will collapse all the observations into a single target, which is useful for targets observed in multiple configurations over multiple projects. 'separate' will make every MS a separate target. I'm not sure this has a particularly clear use case, but may be useful. 

**IMPORTANT** There are some important things here: one is that target definitions shouldn't include mosaic parts -- the code will overwrite this file including those later (it'll pull RA/dec from the observations, and inherit velocity info from the parent). Another is that because this requires overwriting keys, you should restart the `key_handler` after running this to load everything back in.

This is currently a work-in-progress. It works for my test cases (253, 628, 4826 CO(2-1) obs), but may not work in all cases. It does a sanity check that every 'calibration' folder has an associated 'calibrated' folder, which should catch these and we can debug as they come up.

PR #221, opened 2022-05-31
Add exception handling when splitting science SPWs

e-koch:
Wraps the casa split call to catch `RuntimeException`s that occur when a SPW is labelled in the MS (so shows up in listobs, for example) but the table has no rows. This can happen when flag data has been removed for an MS in a previous split.

Specifically, we're encountering this issue for LGLBS VLA data where the OH1612 line is swamped by RFI in some tracks, and so is flagged, but we keep the metadata when splitting so as to keep the same SPW numbering for archival purposes.

### DIFF
--- a/phangsPipeline/casaVisRoutines.py
+++ b/phangsPipeline/casaVisRoutines.py
@@ -274,18 +274,27 @@ def split_science_targets(
             k, split_params[k]) for k in split_params.keys()) +
         ')')
 
-    casaStuff.split(**split_params)
+    # an MS can have a SPW label for data that is no longer contained in the MS
+    # (e.g., it was fully flagged, and keepflags=False was used in a previous split)
+    # This try/except should work for CASA 6 and newer versions with CASA's improved
+    # exception handling.
+    try:
+        casaStuff.split(**split_params)
+        flag_split_success = True
+    except RuntimeError as exc:
+        logger.error("Splitting failed with exception: {}".format(exc))
+        flag_split_success = False
 
     # Re-weight the data if desired.
-
-    if do_statwt:
-        logger.info("Using statwt to re-weight the data.")
-        statwt_params = {'vis': outfile, 'datacolumn': 'DATA'}
-        logger.info(
-            "... running CASA "+'statwt(' +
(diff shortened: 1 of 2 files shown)


e-koch:
The sys.path commit isn't introducing any actually changes here; it just got lost in the many PR merges a few months ago.

e-koch:
This potentially could cause some behaviour changes for which tracks are included/skipped. It _should_ only affect cases where a SPW number corresponds to an empty table, but there may be other cases I'm not thinking of.

I'd be grateful for a quick review from some combo of @1054 @thomaswilliamsastro @astrojysun @low-sky @akleroy before merging in.

They key change is this block: https://github.com/akleroy/phangs_imaging_scripts/pull/221/files#diff-2ee8183b412d753336c37e75f3164c06ac9dc28279b5ea7483430acc4e2eacb5R281

Is this fine to skip over "silently"? Should we be tracking data sets that encounter processing issues?

1054:
The key change looks good!
Maybe just a minor question, in what cases the split will fail? If it's an empty spw issue, maybe we can add a  doublecheck before split? Just in case there are other errors like computer issue or something. Not important thoughts though.

e-koch:
@1054 I _think_ the exception catch will work alright for the cases I have in mind. I'll do a quick test with CASA 5 tomorrow to check it's OK there.

The specific case is when some SPWs have been split out prior to the pipeline (so something listed in the ms file key) using the `reindex=False` with `mstransform`. This keeps the original SPW numbering after the split, but also seems to retain some of the SPW metadata (e.g. if you split out SPW 5, there will be a listing in listobs for SPWs 0-5 even though those visibilities aren't included anymore.

Alternatively, is there a clean check for an empty table selection for an MS without reading the visibilities into memory? We could directly check if the SPW selection gives an empty table (as split is doing and failing on) but I don't want to introduce inadvertent memory issues if trying to read in visibilities to check if the array is empty.

e-koch:
I've now run this on a few different datasets (ALMA, SMA, VLA) and it's only catching this specific case with the VLA data and non-reindexed SPW numbering described above, so I'm going to merge this in.

Issue #223, opened 2022-06-13
Issue with multiply_cube_by_value 


ashleythomasbarnes:
I'm coming across the following issue: 

> /export/data1/shared/densegas_acatp/imaging/ngc2903/imaging_scripts/phangsPipeline/casaCubeRoutines.py in multiply_cube_by_value(infile, value, brightness_unit, huge_cube_workaround)
>     163         myia.putchunk(vals)
>     164
> --> 165     myia.setbrightnessunit(brightness_unit)
>     166     myia.close()
>     167
> UnboundLocalError : local variable 'myia' referenced before assignment

I think its calling the `huge_cube_workaround`, hence should this be moved into the if statement? 

https://github.com/akleroy/phangs_imaging_scripts/blob/69ef31500018a2fa5b4b4d15c28149a21e4db14b/phangsPipeline/casaCubeRoutines.py#L165

e-koch:
@ashleythomasbarnes -- I also hit this and have started a fix that I'll make a PR for later today (which includes a few other things; so someone else should feel free to push a fix in the meantime).

The solution for this part is to call `ia.close()` in the else statement above, and make a new ia tool after the if/else statements before `ia.setbrightnessunit`

1054:
Right! Do not forget to do `myia.open` again.

akleroy:
Might be at least a spiritual cousin of #219

e-koch:
A vague future to-do is to check whether the tools level (`ia`, etc) can be used as context managers so open/close don't need to be tracked everywhere
```
with ia.open('blah') as myia:
     # do stuff with the image
```

e-koch:
Fixed in #224

PR #224, opened 2022-06-14
Update casaCubeRoutines.py

ashleythomasbarnes:
Correction for https://github.com/akleroy/phangs_imaging_scripts/issues/223 
Closed myia in else, and reopend after to add in units to cube.


e-koch:
Confirming that this fixed the same issue for me. Merging in.

Issue #220, opened 2022-05-24
ia.getchunk write by channel workaround


e-koch:
@1054 Here's a possible fix for writing out. For a 3D cube `data`:
```
CASA <10>: for ii,slc in enumerate(data):
      ...:     slc = slc[None,:,:,:]. # Add degenerate stokes dimension
      ...:     ia.putchunk(pixels=slc.T, blc=[-1,-1,-1,ii])
      ...:
```

This may need some switching of axes... but is hopefully somewhat general.

Issue #206, opened 2021-10-14
Add log to file capability


low-sky:
Feature request from @tonywong94

thomaswilliamsastro:
Fixed in #338

PR #225, opened 2022-06-18
Fixing big image putchunk/getchunk/putregion memory issue, adding a smart pipelineLogger

1054:
## Bug fixing: getchunk/putchunk/putregion

Fixing a known issue when dealing with big images, e.g., 2880x2880x393, CASA `image.getchunk()`, `image.putchunk()` or `image.putregion()` breaks with the following error message:
```
2021-08-24 14:53:27     SEVERE  image::putchunk (file binding/tools/image/image_cmpt.cc, line 4354)     Exception Reported: ArrayBase::reform() - total elements differ
```
Our code uses these `getchunk`, `putchunk` and `putregion` commands in `casaCubeRoutines.get_mask`, `casaCubeRoutines.copy_mask`, `casaCubeRoutines.multiply_cube_by_value`, and `casaFeatherRoutines.feather_two_cubes` functions. Therefore, the processing are affected. 

This issue is noticed by us as early as August, 2021 when dealing with high-resolution observations but still with the huge linear mosaic area set by the "keys/overrides.txt" file (forgot to disable it). 
This issue is also noted in [https://casa.nrao.edu/docs/casaref/image.putchunk.html](https://casa.nrao.edu/docs/casaref/image.putchunk.html) as: "_If all the pixels didn’t easily fit in memory, you would iterate through the image chunk by chunk to avoid exhausting virtual memory._"

Therefore, following Eric K's brilliant suggestion #220 , now I made several fixes in "casaCubeRoutines.py" and "casaFeatherRoutines.py". The code first detects if there could be a memory issue by either checking the image size (>=2880x2880x393) or by trying to call `getchunk`. In the case of a memory issue, `getchunk` will return a flattened array instead of an array with the dimensions equaling the image cube. In this way, we set a `no_memory_issue` variable to indicate if it is okay to directly call `putchunk` and `putregion` later, or to use a workaround method. The workaround is iterating the processing channel by channel by specifying the `blc`, `trc` or `region` arguments. 

The new postprocessing part of the code has been tested to be working. 

---

## A better logging module, and testing scripts?

In addition, I am thinking about two improvements: a better logging module that can write to casalog instead of just printing on the screen if in CASA; and some testing scripts. 

I have implemented such a logging module, in "pipelineLogger.py". It aims to replace the existing "phangsLogger.py". It is an override of the `logging.getLoggerClass()` class. Inside its `debug`, `info`, `warning` and `error` functions, it first detects if `casaStuff.casalog` exists, if so, then use `casalog.post()` function to do the logging. If it detects no CASA environment, then simply does what logging.Logger does. Besides, it can log to file as mentioned in Issue #206 .

I have also prepared two preliminary testing scripts, one for the "casaCubeRoutines" and one for the "pipelineLogger". Because our pipeline works in both non-CASA and CASA environments, the testing script should do so as well. To run the `unittest` inside CASA, see the comments at the beginning of these testing scripts. I have run them to verify that my above bug fixing code works. I have tested them under both CASA 5.4.0 and 6.1.1. 

---

## Importing issues, now using try-catch as a workaround

There are also some importing issues in "taskSDintImaging.py":
```
from casatasks.private.cleanhelper import write_tclean_history, get_func_params
```
and in "phangsPipeline/__init__.py":
```
from .handlerAlmaDownload import AlmaDownloadHandler
```
for which I simply use a try-catch workaround. 
I'm not sure how well they are tested before. I would suggest to make some testing scripts to verify if they are working as expected. 

---

## Full testing still on-going

The new pieces of code are also being fully tested end-to-end on the high-res data (image-postprocessing-derived). The full processing has not finished yet, but I'd like to first prepare this PR for review. We can merge a few days later when the full processing is done.

### DIFF
--- a/phangsPipeline/casaCubeRoutines.py
+++ b/phangsPipeline/casaCubeRoutines.py
@@ -26,8 +26,12 @@ from .pipelineVersion import version as pipeVer
 # CASA stuff
 from . import casaStuff
 
-logger = logging.getLogger(__name__)
-logger.setLevel(logging.DEBUG)
+# Logging
+from .pipelineLogger import PipelineLogger
+logger = PipelineLogger(__name__)
+
+#logger = logging.getLogger(__name__)
+#logger.setLevel(logging.DEBUG)
 
 #endregion
 
@@ -121,12 +125,25 @@ def get_mask(infile, huge_cube_workaround=True):
     assert len(cube_shape) in [2, 3, 4]
     
     myia.open(infile)
-    cube_mask = myia.getchunk(getmask=True) # cube_mask.shape = [X, Y, CHANNEL, [STOKES]]
     
-    if np.all(cube_mask.shape == cube_shape): # getchunk was successful, no memory issue
+    no_memory_issue = True
+    cube_mask = None
+    if np.prod(cube_shape) >= 2880*2880*393: # known memory issue
+        no_memory_issue = False
+    if no_memory_issue: # try to see if there is a memory issue
+        # in which case the getchunk will return a flat array instead of the cube shape
+        cube_mask = myia.getchunk(getmask=True) # cube_mask.shape = [X, Y, CHANNEL, [STOKES]]
+        if not np.all(cube_mask.shape == cube_shape): # shape does not match, has memory issue
+            no_memory_issue = False
+    
+    if not huge_cube_workaround: # if workaround is not allowed, go the simplest way
+        no_memory_issue = True
+    
+    if no_memory_issue: # getchunk was successful, no memory issue
         mask = cube_mask
     else:
         # putchunk channel by channel
+        logger.debug('getchunk channel by channel for known memory issue')
         mask = np.full(cube_shape, fill_value=False)
         if len(cube_shape) == 2:
             blc = [0, 0] # [X, Y]
@@ -139,16 +156,19 @@ def get_mask(infile, huge_cube_workaround=True):
                 blc = [0, 0, ichan] # [X, Y, CHANNEL]
                 trc = [-1, -1, ichan] # [X, Y, CHANNEL]
                 cube_mask_per_chan = myia.getchunk(blc, trc, getmask=True)
-                mask[:, :, ichan] = cube_mask_per_chan
+                mask[:, :, ichan] = cube_mask_per_chan[:, :, 0]
         elif len(cube_shape) == 4:
             nstokes = cube_shape[3]
-            nchan = cube_shape[2]
+            nchan = cube_shape[2] # It's okay if Stokes and Spectral axes are swapped.
             for istokes in range(nstokes):
                 for ichan in range(nchan):
                     blc = [0, 0, ichan, istokes] # [X, Y, CHANNEL, STOKES]
                     trc = [-1, -1, ichan, istokes] # [X, Y, CHANNEL, STOKES]
+                    #logger.debug('get_mask blc {} trc {}'.format(blc, trc))
                     cube_mask_per_chan = myia.getchunk(blc, trc, getmask=True)
-                    mask[:, :, ichan, istokes] = cube_mask_per_chan
+                    mask[:, :, ichan, istokes] = cube_mask_per_chan[:, :, 0, 0]
+        else:
+            raise Exception('Could not proceed with cube dimension ' + str(len(cube_shape)))
     
     myia.close()
     
@@ -189,18 +209,65 @@ def copy_mask(infile, outfile, huge_cube_workaround=True):
     assert np.all(mask.shape == out_shape) # input and output shape must be the same
     
     myia.open(outfile)
-    out_mask = myia.getchunk(getmask=True) # out_mask.shape = [X, Y, CHANNEL, [STOKES]]
-    if np.all(out_mask.shape == out_shape): # getchunk was successful, no memory issue
+    
+    no_memory_issue = True
+    out_mask = None
+    if np.prod(out_shape) >= 2880*2880*393: # known memory issue
+        no_memory_issue = False
+    if no_memory_issue: # try to see if there is a memory issue
+        # in which case the getchunk will return a flat array instead of the cube shape
+        out_mask = myia.getchunk(getmask=True) # out_mask.shape = [X, Y, CHANNEL, [STOKES]]
+        if not np.all(out_mask.shape == out_shape): # shape does not match, has memory issue
+            no_memory_issue = False
+    
+    if not huge_cube_workaround: # if workaround is not allowed, go the simplest way
+        no_memory_issue = True
+    
+    if no_memory_issue: # getchunk was successful, no memory issue
         myia.putregion(pixelmask=mask)
     else:
         # putregion channel by channel
+        logger.debug('putregion channel by channel for known memory issue')
         myrg = au.createCasaTool(casaStuff.rgtool)
         nx = out_shape[0]
         ny = out_shape[1]
-        nchan = out_shape[2]
-        for ichan in range(nchan):
-            r1 = myrg.box([0, 0, ichan], [nx-1, ny-1, ichan])
-            myia.putregion(pixelmask=np.take(mask, ichan, axis=2), region=r1)
+        #logger.debug('copy_mask mask.shape {}'.format(mask.shape))
+        if len(out_shape) == 2:
+            blc = [0, 0]
+            trc = [nx-1, ny-1]
+            r1 = myrg.box(blc, trc)
+            #logger.debug('copy_mask putregion blc {} trc {}'.format(blc, trc))
+            myia.putregion(pixelmask=mask, region=r1)
+        elif len(out_shape) == 3:
+            specaxis = 2
+            nchan = out_shape[specaxis]
+            blc = [0, 0, 0]
+            trc = [nx-1, ny-1, 0]
+            for ichan in range(nchan):
+                blc[specaxis] = ichan
+                trc[specaxis] = ichan
+                r1 = myrg.box(blc, trc)
+                #logger.debug('copy_mask putregion blc {} trc {}'.format(blc, trc))
+                myia.putregion(pixelmask=np.take(mask, ichan, axis=specaxis), region=r1)
+        elif len(out_shape) == 4:
+            mycs = myia.coordsys()
+            specaxis = mycs.axiscoordinatetypes().index('Spectral')
+            stokesaxis = mycs.axiscoordinatetypes().index('Stokes')
+            nchan = out_shape[specaxis]
+            nstokes = out_shape[stokesaxis]
+            blc = [0, 0, 0, 0]
+            trc = [nx-1, ny-1, 0, 0]
+            for istokes in range(nstokes):
+                for ichan in range(nchan):
+                    blc[specaxis] = ichan
+                    trc[specaxis] = ichan
+                    blc[stokesaxis] = istokes
+                    trc[stokesaxis] = istokes
+                    r1 = myrg.box(blc, trc)
+                    #logger.debug('copy_mask putregion blc {} trc {}'.format(blc, trc))
+                    myia.putregion(pixelmask=np.take(np.take(mask, blc[3], axis=3), blc[2], axis=2), region=r1)
+        else:
+            raise Exception('Could not proceed with cube dimension ' + str(len(out_shape)))
         myrg.done()
     myia.close()
 
@@ -241,39 +308,52 @@ def multiply_cube_by_value(infile, value, brightness_unit, huge_cube_workaround=
     assert len(cube_shape) in [2, 3, 4]
     
     myia.open(infile)
-    vals = myia.getchunk() # vals.shape = [X, Y, CHANNEL, [STOKES]]
-    if np.all(vals.shape == cube_shape): # getchunk was successful, no memory issue
+    
+    no_memory_issue = True
+    vals = None
+    if np.prod(cube_shape) >= 2880*2880*393: # known memory issue
+        no_memory_issue = False
+    if no_memory_issue: # try to see if there is a memory issue
+        # in which case the getchunk will return a flat array instead of the cube shape
+        vals = myia.getchunk() # vals.shape = [X, Y, CHANNEL, [STOKES]]
+        if not np.all(vals.shape == cube_shape): # shape does not match, has memory issue
+            no_memory_issue = False
+    
+    if not huge_cube_workaround: # if workaround is not allowed, go the simplest way
+        no_memory_issue = True
+    
+    if no_memory_issue: # getchunk was successful, no memory issue
         vals *= value
         myia.putchunk(vals)
     else:
         # putchunk channel by channel
+        logger.debug('putchunk channel by channel for known memory issue')
         if len(cube_shape) == 2:
             blc = [0, 0] # [X, Y]
             trc = [-1, -1] # [X, Y]
-            inc = [cube_shape[0], cube_shape[1]] # [X, Y]
             vals = myia.getchunk(blc, trc)
             vals *= value
-            myia.putchunk(vals, blc, inc)
+            myia.putchunk(vals, blc)
         elif len(cube_shape) == 3:
             nchan = cube_shape[2]
             for ichan in range(nchan):
                 blc = [0, 0, ichan] # [X, Y, CHANNEL]
                 trc = [-1, -1, ichan] # [X, Y, CHANNEL]
-                inc = [cube_shape[0], cube_shape[1], 1] # [X, Y, CHANNEL]
                 vals = myia.getchunk(blc, trc)
                 vals *= value
-                myia.putchunk(vals, blc, inc)
+                myia.putchunk(vals, blc)
         elif len(cube_shape) == 4:
             nstokes = cube_shape[3]
-            nchan = cube_shape[2]
+            nchan = cube_shape[2] # It's okay if Stokes and Spectral axes are swapped.
             for istokes in range(nstokes):
                 for ichan in range(nchan):
                     blc = [0, 0, ichan, istokes] # [X, Y, CHANNEL, STOKES]
                     trc = [-1, -1, ichan, istokes] # [X, Y, CHANNEL, STOKES]
-                    inc = [cube_shape[0], cube_shape[1], 1, 1] # [X, Y, CHANNEL, STOKES]
                     vals = myia.getchunk(blc, trc)
                     vals *= value
-                    myia.putchunk(vals, blc, inc)
+                    myia.putchunk(vals, blc)
+        else:
+            raise Exception('Could not proceed with cube dimension ' + str(len(cube_shape)))
     
     if brightness_unit is not None:
         myia.setbrightnessunit(brightness_unit)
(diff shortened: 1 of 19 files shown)


phangsPipeline/casaCubeRoutines.py
```diff
@@ -120,19 +182,94 @@ def copy_mask(infile, outfile, huge_cube_workaround=True):
...
+    myia.open(outfile)
+    
+    no_memory_issue = True
+    out_mask = None
+    if np.prod(out_shape) >= 2880*2880*393: # known memory issue
+        no_memory_issue = False
+    if no_memory_issue: # try to see if there is a memory issue
+        # in which case the getchunk will return a flat array instead of the cube shape
+        out_mask = myia.getchunk(getmask=True) # out_mask.shape = [X, Y, CHANNEL, [STOKES]]
+        if not np.all(out_mask.shape == out_shape): # shape does not match, has memory issue
```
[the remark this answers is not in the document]
e-koch: Ah I see the note above on the flat array returned. Still curious if anyone has an idea on the above question.

phangsPipeline/casaCubeRoutines.py
```diff
@@ -120,19 +182,94 @@ def copy_mask(infile, outfile, huge_cube_workaround=True):
...
+    #    mask = myia.putregion(pixelmask=mask)
+    #    myia.close()
+    
+    mask = get_mask(infile)
+    
+    # use putregion to update pixel mask
+    
+    myia = au.createCasaTool(casaStuff.iatool)
+    
+    myia.open(outfile)
```
e-koch, line 205: The shape check could also go in a helper function to make it easier to track the ia open/close calls.

phangsPipeline/casaFeatherRoutines.py
```diff
@@ -267,39 +271,108 @@ def feather_two_cubes(
...
-            myia.open(current_interf_file)
-            interf_data = myia.getchunk()
-            interf_data[combined_mask == False] = 0.0
-            myia.putchunk(interf_data)
+        assert np.all(interf_shape == sd_shape)
+        
+        no_memory_issue = True
+        interf_mask = None
+        sd_mask = None
+        if np.prod(interf_shape) >= 2880*2880*393: # known memory issue
```
e-koch, line 295: See other comment on swapping to a helper function for the size check.

phangsPipeline/pipelineLogger.py
```diff
@@ -0,0 +1,178 @@
...
+            if logfile is not None:
+                fileHandler = logging.FileHandler(logfile, mode='a')
+                fileHandler.setFormatter(logging.Formatter(self.file_log_format))
+                self.addHandler(fileHandler)
+                self.file_handler = fileHandler
+    
+    def debug(self, message):
+        if self.hasCasaLog():
+            self.setCasaOrigin()
+            casaStuff.casalog.post(message, 'NORMAL')
```
e-koch, line 70: Does casa have a debug level in the logger?

phangsPipelineTests/test_casaCubeRoutines.py
```diff
@@ -0,0 +1,173 @@
+"""
+How to run this test inside CASA:
```
e-koch, line 2: Awesome! We should setup CI to start running the tests in a future PR.

phangsPipelineTests/test_casaCubeRoutines.py
```diff
@@ -0,0 +1,173 @@
...
+        self.assertTrue(os.path.isdir(outname))
+    
+    def test_get_mask(self):
+        self.set_sys_path()
+        from phangsPipeline import casaCubeRoutines as ccr
+        self.make_test_image('test.fits', overwrite=False)
+        self.import_test_image('test.fits', 'test.image', overwrite=False)
+        self.assertTrue(os.path.isdir('test.image'))
+        ccr.get_mask('test.image')
+        shutil.rmtree('test.image')
```
e-koch, line 125: Is `test.fits` getting cleaned up here, too?

e-koch (changes requested):
This is great @1054! I've added a few comments on clarifying and places to simplify the code a bit. All of these additions are super useful.

A couple general thoughts:
* Is the per-plane read/write much slower than the read at once calls? If it's not a big difference, we could read/write per-plane everywhere assuming that side-steps nearly all of the memory issues, and simplifies the code a lot.
* Adding some unit tests is fantastic. Noting here that we should revisit getting CI running on the repo. This is a big step to eventually getting testing wrapped in here.

phangsPipeline/casaCubeRoutines.py
```diff
@@ -120,19 +182,94 @@ def copy_mask(infile, outfile, huge_cube_workaround=True):
...
+    out_shape = myia.shape() # [X, Y, CHANNEL, [STOKES]]
+    myia.close()
+    
+    assert np.all(mask.shape == out_shape) # input and output shape must be the same
+    
+    myia.open(outfile)
+    
+    no_memory_issue = True
+    out_mask = None
+    if np.prod(out_shape) >= 2880*2880*393: # known memory issue
```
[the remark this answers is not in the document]
low-sky: Is this threshold for a specific hardware configuration?  It _looks_ like a magic number for a given piece of hardware.

low-sky (commented):
I think the general question of "why not just operate planewise on the getchunk should be addressed?  It would be good to know that all the if-then-else stuff is needed rather than the single simple approach.  No other comments.

phangsPipeline/casaCubeRoutines.py
```diff
@@ -85,33 +89,91 @@ def get_mask(infile, huge_cube_workaround=True):
...
-        myia = au.createCasaTool(casaStuff.iatool)
-        myia.open(infile+'.temp_deg')
-        mask = myia.getchunk(getmask=True)
-        myia.close()
-
-    os.system('rm -rf ' + infile + '.temp_deg')
-
+        # putchunk channel by channel
+        logger.debug('getchunk channel by channel for known memory issue')
+        mask = np.full(cube_shape, fill_value=False)
```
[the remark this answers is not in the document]
1054: Yes, the `False` is the value to be filled (fill_value is not a boolean flag but takes values), so the output is exactly a boolean array.

phangsPipeline/casaCubeRoutines.py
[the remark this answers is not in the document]
1054: This `2880*2880*393` is the known case where the memory issue happens. I'm adding this check to make the code run faster for such known issues. Otherwise it will waste several to tens of minutes to try to run `getchunk` then figure out there is a memory issue. It's not so friendly for code readers indeed, but I think helps to make the code efficient. Not sure if my idea is good or bad though.

phangsPipeline/casaCubeRoutines.py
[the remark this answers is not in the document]
1054: CASA will not raise an exception during the `getchunk`, but later when calling `putchunk` or `putregion`, it directly crashes..

phangsPipeline/casaFeatherRoutines.py
```diff
@@ -267,39 +271,108 @@ def feather_two_cubes(
...
+            
+                myia.open(current_sd_file)
+                sd_data = myia.getchunk()
+                sd_data[combined_mask == False] = 0.0
+                myia.putchunk(sd_data)
+                myia.close()
+        
+        else:
+            
+            assert len(interf_shape) >= 3
```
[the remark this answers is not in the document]
1054: Right, indeed 4D is not handled well here. Needs some revision here. I'll try to update it soon.

phangsPipeline/pipelineLogger.py
1054: Right, it should be set to `'DEBUGGING'` (or `DEBUG1` or `DEBUG2`), according to [https://casa.nrao.edu/casadocs/casa-6.0/usingcasa/casa-logger](https://casa.nrao.edu/casadocs/casa-6.0/usingcasa/casa-logger).

phangsPipeline/pipelineLogger.py
```diff
@@ -0,0 +1,178 @@
...
+    
+    def error(self, message):
+        if self.hasCasaLog():
+            self.setCasaOrigin()
+            casaStuff.casalog.post(message, 'SEVERE')
+            self.restoreCasaOrigin()
+        else:
+            super(PipelineLogger, self).error(message)
+
+    def findCallerPy37(self, stack_info=False):
```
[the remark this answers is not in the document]
1054: Ah, forgot to comment out `findCallerPy37`, it is not used at all.

phangsPipelineTests/test_casaCubeRoutines.py
1054: No actually, another to-do item indeed.

1054:
Minor revisions done for dealing with 4D cube, debug mode, and testing tearDown. 
Maybe still need to decide (but not a stopper for merging): 

- whether we always want to run `getchunk`, `putchunk` and `putregion` per channel. If so, we do not need the `no_memory_issue` check.
- whether we want to add a helper function to test the `no_memory_issue` flag, or just leave the current code as it is.

e-koch:
I'm not opposed to merging the current form. @1054 I think it's worth a final pass cleaning up the commented out code (e.g., the unused py37 stuff in the logging, and the pyfits code).

The helper function for the memory check is a good idea just to have a single line with the "magic" size number. Basically just:
```
def is_huge(shape):
    return np.prod(shape) >= 2880*2880*393
```

e-koch:
@1054 -- The always loop through channel approach came up in another context with @low-sky and @akleroy. We're thinking there's pretty good utility in just using the per channel approach in all cases to (in practice) never have to hit the memory limit. The time added in the extra for loop must be pretty small compared to most of the other stages in post-processing.

Let us know if you want some testing or a hand with the further changes!

thomaswilliamsastro:
> 



> @1054 do you see any issues we're not thinking of for using the per plane approach everywhere?

I ran into this a bit with the sdintimaging, since it does a per-plane feather -- in that case, it's *significantly* slower doing it per-plane than all at once (at least order of magnitude slower), so I'm not sure this is always the best approach, but for big cubes I think falling to this makes sense

1054:
Yes, thanks a lot for the above discussions! I think doing some tests are really needed to understand the time expense. Thinking about some tests now...

1054:
I finally tested whether we can set the per-channel getchunk/putchunk as the default behavior or not. The answer is no. It's nearly 1.6x slower. 

In detail, I tested the "casaFeatherRoutines.py" 'feather_two_cubes' function with a real data set with a dimension of (231, 232, 178) [X, Y, channel], finding that
- running 'feather_two_cubes' without per-channel getchunk/putchunk:
  - Ran 2 tests in 42.220s
  - Ran 2 tests in 40.124s
  - Ran 2 tests in 43.081s
- running 'feather_two_cubes' *without* per-channel getchunk/putchunk:
  - Ran 2 tests in 76.737s
  - Ran 2 tests in 72.826s
  - Ran 2 tests in 74.754s

I wrote a unittest script for this test: "phangsPipelineTests/test_casaFeatherRoutines.py". To reproduce this test, we first need to prepare the test data under the "phangsPipelineTests/test_data" directory, then cd into the "phange_imaging_script" directory, then launch CASA and run following commands:
```
sys.path.append('../analysis_scripts') # make sure we can import analysisUtils as au
sys.path.append('.') # make sure we can import phangsPipelineTests
import phangsPipelineTests
phangsPipelineTests.TestingCasaFeatherRoutinesInCasa().run()
```

As for the coding, now a more generalized function is added into "casaCubeRoutines.py", named 'check_getchunk_putchunk_memory_issue', to check the memory issue. It is used in "casaCubeRoutines.py" (by 'get_mask', 'copy_mask' and 'multiply_cube_by_value' functions) and "casaFeatherRoutines.py" (by 'feather_two_cubes' function). 

This should basically solve all the above to-do items.

phangsPipelineTests/test_casaCubeRoutines.py
1054: now has 'unittest.TestCase.tearDown'

phangsPipeline/casaCubeRoutines.py
```diff
@@ -85,33 +89,91 @@ def get_mask(infile, huge_cube_workaround=True):
...
+    cube_mask = None
+    if np.prod(cube_shape) >= 2880*2880*393: # known memory issue
+        no_memory_issue = False
+    if no_memory_issue: # try to see if there is a memory issue
+        # in which case the getchunk will return a flat array instead of the cube shape
+        cube_mask = myia.getchunk(getmask=True) # cube_mask.shape = [X, Y, CHANNEL, [STOKES]]
+        if not np.all(cube_mask.shape == cube_shape): # shape does not match, has memory issue
+            no_memory_issue = False
+    
+    if not huge_cube_workaround: # if workaround is not allowed, go the simplest way
```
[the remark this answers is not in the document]
1054: we can probably leave the huge_cube_workaround there for compatibility.. can remove later if we're sure this will not break anything..

1054 (commented):
Implemented unittest script "phangsPipelineTests/test_casaFeatherRoutines.py". 
Tested that switching to per-channel getchunk/putchunk will be much slower. 
Implemented the common/helper function 'check_getchunk_putchunk_memory_issue' in "casaCubeRoutines.py".

phangsPipelineTests/test_casaCubeRoutines.py
1054: Done

1054:
Screenshot 2022-09-09 at 02 24 40

Any idea how to mark this "1 change requested" done?

1054:
Adding a small but important feature: allowing user to override imaging size by setting `deltara` `deltadec` in the "overrides.txt" file. 
Also adding several unittest scripts.

phangsPipeline/casaCubeRoutines.py
```diff
@@ -26,11 +26,103 @@
...
+#region Check getchunk putchunk memory issue
+
+def check_getchunk_putchunk_memory_issue(
+    infile, 
+    myia=None,
+    return_myia=False,
+    return_data=False,
+    return_mask=False,
+    return_shape=False,
+    huge_cube_workaround=True,
```
e-koch, line 47: I'd vote to just drop the `huge_cube_workaround` kwarg as your changes here negate needing it.

phangsPipeline/casaCubeRoutines.py
```diff
@@ -85,33 +177,74 @@ def get_mask(infile, huge_cube_workaround=True):
...
-
-    else:
-        myia = au.createCasaTool(casaStuff.iatool)
-        myia.open(infile+'.temp_deg')
-        mask = myia.getchunk(getmask=True)
-        myia.close()
-
-    os.system('rm -rf ' + infile + '.temp_deg')
-
+    #if huge_cube_workaround:
```
e-koch, line 180: We should clean up the unused `huge_cube_workaround` code throughout.

phangsPipeline/casaCubeRoutines.py
```diff
@@ -26,11 +26,103 @@
...
+        has_memory_issue = True
+    if not has_memory_issue: # try to see if there is a memory issue
+        # in which case the getchunk will return a flat array instead of the cube shape
+        if return_data or not return_mask:
+            cube_data = myia.getchunk() # data.shape = [X, Y, CHANNEL, [STOKES]]
+            check_shape = cube_data.shape
+        if return_mask:
+            cube_mask = myia.getchunk(getmask=True) # mask.shape = [X, Y, CHANNEL, [STOKES]]
+            check_shape = cube_mask.shape
+        if not np.all(check_shape == cube_shape): # shape does not match, has memory issue
```
e-koch, line 101: I didn't fully follow why this function can also return the whole cube or mask. Is this the key line? That we need to be able to fit the mask or data into memory to have the correct shape?

phangsPipelineTests/test_handlerImaging.py
```diff
@@ -0,0 +1,112 @@
...
+        this_kh = kh.KeyHandler(master_key=self.test_keys_dir+os.sep+'master_key.txt')
+        this_imh = imh.ImagingHandler(key_handler=this_kh)
+        this_imh.loop_imaging(
+            do_all=True,
+            )
+        os.chdir(self.current_dir)
+    
+    def tearDown(self):
+        os.chdir(self.current_dir)
+        #for this_dir in ['cleanmasks', 'reduction']:
```
e-koch, line 87: Add this back in for clean up?

phangsPipelineTests/test_handlerVis.py
```diff
@@ -0,0 +1,92 @@
...
+        this_uvh = uvh.VisHandler(key_handler=this_kh)
+        this_uvh.loop_stage_uvdata(
+            do_all=True,
+            )
+        os.chdir(self.current_dir)
+        assert os.path.isdir(os.path.join(self.test_data_dir, 'reduction', 'imaging', 'ngc4321', 'ngc4321_7m_co10.ms'))
+        
+    def tearDown(self):
+        os.chdir(self.current_dir)
+        #for this_dir in ['cleanmasks', 'reduction']:
```
e-koch, line 67: Also here. Clean up for the intermediate vis files?

e-koch (approved):
@1054 -- just a couple minor questions I had, but otherwise I think this looks great and we should merge soon.

One additional request: can you add to the README on how to run the test suite?

phangsPipeline/casaCubeRoutines.py
1054: This is mainly for the efficiency. We will anyway run a `getchunk` to test if it returns a correct array size or not. I let the code to run `getchunk()` with getmask=True if return_mask, or getmask=False if return_data, so that if it is executed successfully, we can immediately return the result as the user wanted. If not doing so, we will need to rerun getchunk in the later part of the code to get what we want in `casaCubeRoutines.get_mask`, `casaCubeRoutines.copy_mask`, `casaCubeRoutines.multiply_cube_by_value`, and `casaFeatherRoutines.feather_two_cubes`, and waste several to tens of minutes. 

In the case that there is a memory issue for big data array, the code will set has_memory_issue=True, and will set the returned data or mask to None even if return_data or return_mask has been set. The later part of the code will check has_memory_issue and do per-channel computations correspondingly.

phangsPipeline/casaCubeRoutines.py
1054: Would this break Tom's scripts for some earlier data reduction, or other parts of the pipeline if there is a `huge_cube_workaround` leftover? For compatible reason I tend to keep it...

1054:
Adding a function in the ccr to trim the edge channels which have a much coarser beam due to the merging of datasets that have non-identical frequency setups. This fixes the issue that high-res cube processing got a too large common beam due to these edge channels.

e-koch:
@1054 -- do you feel this is ready to merge? I feel we should merge all of these great changes in soon

1054:
> @1054 -- do you feel this is ready to merge? I feel we should merge all of these great changes in soon

Oh sorry for my late reply! Yes, I think it has no problem to merge, if no conflicts  ;)

PR #228, opened 2022-08-30
Chunked cube imaging

e-koch:
This PR adds a handler to split the imaging of cubes into smaller chunks (e.g., image individual channels). It's meant primarily for imaging really large data cubes that either hit memory limits on normal machines or take a very long time to run (>1 week). While it could be run on a single machine, I'm writing the new handler with distributed computing in mind, where a "chunk" of a cube is imaged as a single job on the cluster.

* [x] Alter the clean call creation to understand/create sub-clean jobs based on N number of chunks
* [x] Task to split MS by chunks
* [x] Task to initialized per chunk clean calls
* [ ] Task to setup and split SD data for SDintimaging
* [x] Task for dirty imaging
* [x] Task for MS imaging
* [x] Task for SS masking
* [x] Task for SS imaging
* [x] Task for gathering chunks into a cube
* [x] Task for splitting intermediate products from a cube to chunks
* [ ] Task to split user clean masks per chunk
* [x] Task to (optionally) export chunks to FITS
* [ ] Double-check link ups to various SDintimaging code (names, alternate handling, etc)

Tests:
* [x] Local small cube test for consistency split by different chunk sizes
* [x] Small cube test of normal vs. chunked imaging for consistency
* [x] Full test with 1 chunk as test for job submission on a cluster
* [ ] Full test run with SDintimaging


e-koch:
I've resurrected this long-standing PR based on the newly released [CASA Memo 13](https://drive.google.com/file/d/1_8JeN-MtDEqUYjRn7eIUbqcE0EyJeSqW/view) led by @Kitchi which demonstrates (among other things) that _there is no performance gain in splitting channel to independent MSs_ for parallelized tclean runs.

This avoids several difficulties in tracking and producing an additional set of MSs (for which I was running into an apparent bug that ended up being nearly impossible to reproduce where MS metadata would lead the last channel in an image to be flagged).

The current version in this PR implements a separate image handling/execution class `ImagingChunkedHandler` that reproduces the PHANGS imaging recipe in sub-cubes. However, it will _not_ work for separately imaging each channel as this breaks the noise and masking routines.

e-koch:
See benchmarking scripts here: https://github.com/Kitchi/cube_parallelization_benchmarking

e-koch:
Basic functionality is working based on a local test.

A few to-dos remain above but this is minimally working in its current form.

e-koch:
Pinged for a few reviews while I finish the last checks. Thanks to anyone who has time to look through!

run_casa_imaging_chunked_example.py
```diff
@@ -0,0 +1,197 @@
...
+# the imaging loop call but this call does everything.
+
+if do_imaging:
+
+    # Initialize the ImagingChunkedHandler
+    # Unlike ImageHandler, you must specify the target, configuration and line_name
+    # ImagingChunkedHandler is designed to image individual products that require
+    # chunking to process efficiently.
+    this_imh = ImagingChunkedHandler('ic10ctr', 'C', 'hilores', this_kh,
+                                     chunksize=10,
```
low-sky, line 167: Maybe add some guidance on chunksize relative to available RAM?

e-koch: Good point. I'll need to do that upstream.

Right now, this should still default to triggering tclean's subcube mode.

phangsPipeline/handlerImagingChunked.py
```diff
@@ -0,0 +1,1521 @@
...
+                for ii, chunk_num in enumerate(chunks_iter):
+
+                    # Make the chunk clean call:
+                    this_clean_call = self.task_initialize_clean_call(chunk_num, stage='dirty')
+
+                    logger.info("")
+                    logger.info("&%&%&%&%&%&%&%&%&%&%&%&%&%")
+                    logger.info("Making dirty image for chunk {}:".format(chunk_num))
+                    logger.info(str(this_clean_call.get_param('imagename')))
+                    logger.info("This is {0} out of {1} to be imaged".format(ii, len(chunks_iter)))
```
low-sky, line 842: zero indexing makes this info off by one.  Does it matter? Probably not.

phangsPipeline/clean_call.py
```diff
@@ -200,12 +208,58 @@ def get_param(self, key=None):
...
+        ----------
+        chunksize : int
+            Number of channels to image together in a chunk.
+        '''
+
+        if not self.use_chunks:
+            logger.warn("use_chunks has not been enabled. Call `CleanCall` with `use_chunks=True`.")
+            return
+
+        if chunksize < 1:
```
low-sky, line 225: Add an assert `int` for chunksize?

phangsPipeline/handlerImagingChunked.py
```diff
@@ -0,0 +1,1521 @@
...
+                    else:
+                        logger.warn("Cube name {} already exists. Skipping.".format(fname_dict[img_type]) +
+                                    " Enable overwrite=True to force writing a new version.")
+                        continue
+
+                logger.info("Making {0} from this list of chunked images:".format(fname_dict[img_type]))
+                logger.info(chunk_fname_dict[img_type])
+
+                ia = casaStuff.image()
+                im = ia.imageconcat(outfile=fname_dict[img_type],
```
low-sky, line 788: I took a scan through and tried to follow but it seems like there's a risk of the end_chan from one chunk being the same as the start_chan for the next and being imaged  I don't see exactly where this is dropped or if there is an overlap just from the code alone.  Anyways, this whole PR doesn't seem like it would work if there's actually a problem at all so I think it's just a case of I'm not seeing where it's bookkept.

low-sky (approved):
Seems solid.  Enough to get into broader testing.  Thanks!

phangsPipeline/handlerImagingChunked.py
e-koch: That check gets handled in `CleanCall.return_chunked_channel_ranges` in this line: https://github.com/akleroy/phangs_imaging_scripts/pull/228/files#diff-b44a497b1b87cd2eae0ee857b74cf7453d91a88f47f54ac13f931c2649b59126R253

(This was actually an issue that kept cropping up when splitting the MS up, but I'm not getting any issues when running the chunks from the original whole MS).

phangsPipeline/clean_call.py
e-koch: fixed!

phangsPipeline/handlerImagingChunked.py
e-koch: No but certainly minimizes confusion. fixed!

bbtufty (approved):
Looks good to me!

phangsPipeline/__init__.py
```diff
@@ -12,6 +12,8 @@
 from .phangsLogger import setup_logger
 from .handlerKeys import KeyHandler
 from .handlerSingleDish import SingleDishHandler
+# TODO: REMOVE THIS COMMENT AFTER COMPLETION
```
thomaswilliamsastro, line 15: Flagging this for future

thomaswilliamsastro (approved):
Looks good to me!

phangsPipeline/__init__.py
e-koch: fixed!

There's now an empty `AlmaDownloadHandler` defined when any of the other dependencies are missing.

e-koch:
Ok, this can be merged in its current form to make initial testing easier. Non of the changed here should alter the normal pipeline steps.

There's now a couple examples of deploying chunked imaging and individual targets to a cluster to run in parallel (as we're definitely going to increasingly hit cases where this makes imaging easier).

The remaining steps to be addressed in an upcoming PR are:
- Ensuring clean masks are appropriately split to match the chunked imaging jobs
- Finish connecting the additional steps to use SDIntimaging with the chunked imaging handler

PR #230, opened 2022-11-22
Allow setting fixed frequency ranges to avoid in uvcontsub

e-koch:
Adds an additional keyword entry for line definitions in `config_definitions.txt`:  

`exclude_freq_ranges_ghz`: given as a list of list, this gives a set [low, high] frequency range in GHz to avoid when using uvcontsub.
For example:
```
line_product    hi21cm    {'line_tag':'hi21cm','channel_kms':0.42,'statwt_edge_kms':50.0}
line_product    hi21cm    {'fitorder':0,'combinespw':False,'lines_to_flag':['hi21cm'],'exclude_freq_ranges_ghz':[[1.42002671,1.42078479]]}
```

This is useful when multiple targets emit in the same transition along a line-of-sight (e.g., Galactic and extragalactic 21-cm HI). The example above sets a fixed v_rad+/-80 km/s avoidance window for Galactic HI.

This appears to working well for our current HI data, and should have no impact on other uses as the keyword is ignored when not present.

FYI @akleroy

### DIFF
--- a/phangs-alma_keys/config_definitions.txt
+++ b/phangs-alma_keys/config_definitions.txt
@@ -6,14 +6,14 @@
 # "configurations" that to be processed and combined by the
 # pipeline. Here, the user provides detailed parameters for each
 # product and config. The file has three columns:
- 
+
 # Column 1: type of field (see below)
 
 # Column 2: name (this can be anything, it will appear in file names
 # and be used by the handlers and keys)
 
 # Column 3: parameters (write as literal dictionaries with no spaces)
- 
+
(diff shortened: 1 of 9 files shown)


e-koch:
@thomaswilliamsastro mind having a quick look through? This is working in production for the LGLBS HI imaging.

I need to merge this soon for the next round of LGLBS processing, though.

e-koch:
I'm encountering no issues with my updates to the imports. Please ping if anyone does hit a problem related to these changes!

PR #234, opened 2023-04-11
adjusted imports and small key reader patch

akleroy:
@1054 @e-koch or @low-sky 

See if this makes sense - I adjusted the imports in casaStuff and then executed it in CASA versions from 4.5 through 6.5 . I think it should now not trigger errors, import successfully and be more stable. It's a beast though.

I also made a small key reader patch for the distance key. I'm not sure what happened, but re.match might have changed across python version. It works in the new form as far as I can tell.

### DIFF
--- a/phangsPipeline/casaStuff.py
+++ b/phangsPipeline/casaStuff.py
@@ -1,9 +1,30 @@
 # CASA imports
-try:
-    from taskinit import *
 
-    # Import specific CASA tasks. Not all of these are used by this
-    # package, so this could be pared in the future.
+# This is a huge pain. Check that it works correctly by running
+
+# casapy-XYZ -c casaStuff.py --nologger
+
+# with XYZ each relevant version.
+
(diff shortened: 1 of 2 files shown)


akleroy:
wait - ugh. Hold on. Will reissue in a bit.

akleroy:
OK now it's good. I think.

e-koch (commented):
Added a note on the `dist_mpc` check. I think this looks good.

@1054 can you do the final check and merge?

phangsPipeline/casaStuff.py
1054, line 1: Since `almahelpers_localcopy` is a module of the `analysis_scripts`, do we want to use a try and warning structure?
```
if casa_version[0] <= 5:
    from recipes.almahelpers import tsysspwmap
else:
    #from almatasks.private.almahelpers import tsysspwmap
    try:
        from almahelpers_localcopy import tsysspwmap
    except ModuleNotFoundError:
        print('Warning! CASA analysis_script module "almahelpers_localcopy" is not found. Single dish reduction will be unavailable, but interferometry and feathering will not be affected.')
```

PR #238, opened 2023-06-08
pip CASA updates

thomaswilliamsastro:
Some tidy-up to get the pipeline working in the latest pip-installed CASA environment. Retains backwards compatibility for earlier versions. Currently a draft until I've managed to chunk something through successfully, but I have got through imaging at least

* Add ReleaseHandler to __init__.py
* Various import fixes for newer python/CASA versions
* uvcontsub -> uvcontsub_old for new CASA versions
* Fix 'if' statements for == instead of 'is'
* Remove deprecated np.float calls
* Replace 'clobber' with 'overwrite' for hdu.writeto() calls if using astropy

### DIFF
--- a/phangsPipeline/casaStuff.py
+++ b/phangsPipeline/casaStuff.py
@@ -10,6 +10,8 @@
 
 # Obtain a version tuple (note the syntax change from < 6 to > 6)
 
+from packaging import version
+
 try:
     from taskinit import *
 except ModuleNotFoundError:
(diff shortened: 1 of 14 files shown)


phangsPipeline/casaStuff.py
```diff
@@ -107,7 +109,11 @@
                            tclean,
                            uvcontsub,
                            visstat)
-    
+
+    # TODO: For now, uvcontsub doesn't work as we want it in newer CASA versions, fall back to old version
+    if casa_version_str >= '6.5.2':
```
low-sky, line 114: Does this work?  I'd figure it would be something like:
```
from packaging import version
version.parse(casa_version_str) >= version.parse('6.5.2')
```

low-sky (approved):
Provided version comparison works as intended, I think this is fine.  However, I note that 
'6.5.2' > '6.11.4'  is `True` so I think strict string comparison isn't right here.

phangsPipeline/casaStuff.py
thomaswilliamsastro: You're right, pushed this as an update. I've seen it just done in strings before so figured it'd be good, but I guess not

thomaswilliamsastro:
I'll note here that the releaseHandler stuff needs some more heavy-duty refactoring, since it's calling keyHandler functions that no longer exist. I've managed to at least sort the other package deprecations but that would need to be another PR

PR #241, opened 2024-06-05
Handle empty masks for Tpeak

e-koch:
`write_tmax` fails when the mask is empty in more recent versions of numpy (as `np.min/max` now raises an error when passed an empty array).


e-koch:
This fixed things for my local test. I strongly suspect this can be merged with no unintended failures resulting.

low-sky (approved):

PR #244, opened 2024-09-13
Allow small freqtol in concat for lines

e-koch:
Implement method for allowing concatenating SPWs with CtrFreq within a fraction of the channel width.


akleroy (approved):
Seems sensible to me. Merge?

PR #245, opened 2024-09-18
Cleanup chunked imaging

e-koch:


akleroy:
I'm doing it - we're effectively developing on main. Can't stop me!

e-koch:
**Use task_cleanup with caution!! I didn't finish an initial local test and there's an rm -rf in there!**

PR #246, opened 2024-10-09
Temp imaging directories for chunked imaging handling

e-koch:
Implement temp dirs when imaging in chunks to avoid overlapping casa temp products. 

I'm hitting issues with the filesystem I'm running this on where the casa temporary gridding and weight products are interacting between the separate jobs running each chunk (since the processing is all in the same directory).


e-koch:
Update: the temp directory fixes various issues with running multiple CASA instances in the same directory.

thomaswilliamsastro (approved):

e-koch:
Just to note: this is very useful for cluster processing and what we should adopt whenever running multiple jobs in parallel.

PR #251, opened 2025-01-23
Add shuffled cubes and flat maps scripts

lukas-neumann-astro:


scripts/.DS_Store
e-koch, line 1: Can you delete this?

e-koch (commented):
Looks good! Just remove the .DS_Store file that got pushed.

e-koch (approved):

PR #252, opened 2025-02-13
uvcontsub update

e-koch:
Adds handling for the new uvcontsub task inputs (CASA >=6.5.2). Though note that some options are no longer available to the new task (e.g. spw combination).

Also adds a new line config input parameter `flag_edge_fraction` that avoids using a fraction of the channels near the edge of a SPW. This seems to improve contsub fitting for >4 MHz EVLA HI SPWs. Default is 0.0 so will otherwise have no effect from previous default behaviour.


e-koch:
Don't merge this yet. I'm still testing locally.

e-koch:
Overdue update: this worked fine for the LGLBS cases where we were encountering contsub issues.

thomaswilliamsastro (approved):
Good to see this actually implemented now, rather than my hack to just use the old one...just for my benefit, have you run any tests just to see if this new task produces the same results as the old uvcontsub? It'd be good to know if anything changes and if so, by how much

e-koch:
No, I haven't compared to the old uvcontsub outputs. though there's not much we can do about discrepancies and those should be captured upstream in CASA testing. (which doesn't mean we shouldn't compare for a few cases!)

PR #256, opened 2025-05-06
Updated TP pipeline

1054:
The updated TP pipeline with many fixes, compatible with CASA 5 and the pipeline version of CASA 6, can handle multiple EBs and multiple line config.


akleroy:
I'd like to merge this - I haven't run it but it addresses current issues I've run into and the changes look reasonable.

Currently the file handling and image size estimation in the imaging step is crashing frequently - @1054 's changes fix the file handling and seem like they at least try to fix the image size estimation?

PR #259, opened 2025-08-21
Masking updates

akleroy:
(1) Adds cleaner file and i/o handling to velocity field based masking and spectral stacking.

(2) Documentation of the stacking module (first two routines) with some variable name shifts (sorry).

(3) Fixed a bug in the stacking code where the mask was not being applied after shifting: specifically now use the shifted mask (before wasn't doing this) and now have the criteria set at > 0.5 (before > 0.0 which produced issues).

### DIFF
--- a/phangsPipeline/scMaskingRoutines.py
+++ b/phangsPipeline/scMaskingRoutines.py
@@ -394,7 +394,7 @@ def make_vfield_mask(cube_in, vfield_in, window_in,
 
     -----------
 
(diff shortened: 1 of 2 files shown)


akleroy:
It occurs to me that the Shuffle should also add NaNs when shifting outside the velocity range. The FFT will roll things. This PR can go in once it seems okay to one person, but I'll also add that when I get a minute at some point.

PR #262, opened 2025-09-02
Noise prediction generation

akleroy:
This checks in some image handling stuff to support generation of predicted noise maps or "time-like" exposure maps from lists of fields and primary beams. I exercised the test case for the 10 pc survey of the AOT-generated field centers. 

I'd like to expand to P2G-pointing list (with Erik's tips) and a listobs or ms-tool generated list but that's not the immediate use case. So checking this in now.

No obvious conflicts with anything, so probably just needs one reviewer to give it the "sanity" check.

### DIFF
--- a/phangsPipeline/utilsImages.py
+++ b/phangsPipeline/utilsImages.py
@@ -7,6 +7,8 @@ import warnings
 import math
 
 import astropy.units as u
+from astropy.units import Quantity
+
 from astropy.io import fits
 from astropy.wcs import WCS
 from astropy.coordinates import SkyCoord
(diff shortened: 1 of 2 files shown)


phangsPipeline/scTheoreticalNoise.py
```diff
@@ -0,0 +1,201 @@
...
+            (offset_arcsec <= support_factor*beam_fwhm_arcsec)
+
+        #plt.imshow(ra_img, origin='lower')
+        #plt.show()
+        
+        #plt.imshow(offset_arcsec, origin='lower')
+        #plt.show()
+
+        #plt.imshow(this_exposure_map, origin='lower')
+        #plt.show()
```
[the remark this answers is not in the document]
e-koch: Or kwarg option set to False by default.

phangsPipeline/scTheoreticalNoise.py
```diff
@@ -0,0 +1,201 @@
...
+import astropy.wcs as wcs
+import astropy.units as u
+from astropy.convolution import convolve, Gaussian2DKernel
+from astropy.io import fits
+from astropy.stats import mad_std
+from astropy.table import Table
+
+from matplotlib import pyplot as plt
+
+from spectral_cube import SpectralCube
```
e-koch, line 21: For the filename and this import, you're not actually using spectral-cube here. Suggest `utilsTheoreticalNoise` instead for consistency.

phangsPipeline/scTheoreticalNoise.py
```diff
@@ -0,0 +1,201 @@
...
+    exposure_map = np.zeros_like(ra_img, dtype=float)
+
+    # Beam size in sigma units
+    beam_sigma_arcsec = beam_fwhm_arcsec / 2.355
+
+    for this_ra, this_dec, this_weight in \
+        zip(field_ra, field_dec, weight):
+        
+        # Assume Euclidean
+        offset_arcsec = np.sqrt(np.cos(this_dec/180.*np.pi)**2*(ra_img - this_ra)**2 + \
```
e-koch, line 167: since you have astropy.units support imported, consider switching to `* u.deg).to(u.arcsec)`.

But this is minor/pedantic and you can ignore.

phangsPipeline/utilsImages.py
```diff
@@ -0,0 +1,160 @@
...
+        brc_coord = SkyCoord(
+            brc_coord[0]*u.deg, brc_coord[1]*u.deg, frame='icrs')
+
+    # Find the midpoint
+    pa = tlc_coord.position_angle(brc_coord)
+    sep = tlc_coord.separation(brc_coord)
+    midpoint = tlc_coord.directional_offset_by(pa, sep/2.0)
+    
+    # Find the edge
+    extent_deg = sep.data/np.sqrt(2)
```
e-koch, line 43: Same pedantic point above: switch to unit conversions with astropy.units.

phangsPipeline/utilsImages.py
```diff
@@ -0,0 +1,160 @@
...
+from astropy.wcs import WCS
+from astropy.coordinates import SkyCoord
+
+# ------------------------------------------------------------------------
+# Make new headers
+# ------------------------------------------------------------------------
+
+def make_simple_header_from_box(
+        tlc_coord = None, brc_coord = None,
+        pix_scale_deg = None, pad_arcsec = 0.0):
```
e-koch, line 20: Suggest passing these as angular quantities rather than having the kwarg define the unit.

for comparisons,
```
pix_scale = pix_scale.to(u.deg)
```
and allow astropy's builtin unit equivalence checks to kick in.

phangsPipeline/utilsImages.py
```diff
@@ -0,0 +1,160 @@
...
+    # Deal with center, skycoords, units, etc. better
+    
+    if nx is not None and ny is not None:
+        extent_x = pix_scale * nx
+        extent_y = pix_scale * ny
+    elif extent_x is not None and extent_y is not None:
+        nx = int(np.ceil(extent_x*0.5 / pix_scale) * 2 + 1)
+        ny = int(np.ceil(extent_y*0.5 / pix_scale) * 2 + 1)
+
+    new_hdr = fits.Header()
```
e-koch, line 83: No issues but reading into the broad public record that having pre-defined "typical" headers or WCS objects be defined would be useful!

e-koch (approved):
Added a few things that may be useful in the future but otherwise aren't hard issues that need to be dealt with ASAP.

PR #277, opened 2025-12-09
Allow field specification in TP pipeline

thomaswilliamsastro:
- Allow field specification in TP pipeline


e-koch (approved):
Looks good to me. We should revisit when revamping the SD pipeline for any config key changes.

PR #280, opened 2025-12-15
SingleDish pipeline can now handle multiple MS's

thomaswilliamsastro:
- SingleDish pipeline can now handle multiple MS's

Also a little general code tidy-up, to improve readability


thomaswilliamsastro:
Only other thing I changed (and maybe it's not necessary) is for the imaging it'll pull the mean frequency from the concatenated MS, rather than the first one concatted. There might be a potential crash if `phcenter` is not defined in that function call, but from the pipeline logic it's always passed along so I didn't see it come up

e-koch (approved):

e-koch:
@thomaswilliamsastro good to merge this on? I'll need to handle rebasing for #281

PR #282, opened 2026-01-12
Accelerate noise generation

low-sky:
Speedups for noise generation.  Profiling suggests the big problem was in the convolution being "not FFT" rather than in stride issues but these are implemented anyways.

### DIFF
--- a/phangsPipeline/scNoiseRoutines.py
+++ b/phangsPipeline/scNoiseRoutines.py
@@ -7,7 +7,7 @@ import scipy.stats as ss
 from scipy.signal import savgol_coeffs
 import astropy.wcs as wcs
 import astropy.units as u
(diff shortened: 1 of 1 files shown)


thomaswilliamsastro:
This makes me happy. Do you have any metrics on how much faster this ends up being?

low-sky:
Immense improvements for heavily oversampled beams (>factor of 10).  For reasonable pixels per beam, it's a factor of a few.  

There are some issues with the heavily oversampled beams and the largest scales of convolution.  There's clearly not good spectral sampling in this case and there's signal in the noise.  That's a separate issue so this can be looked at now.

thomaswilliamsastro:
Looks good to me, I've squashed/rebased and done a little tidy up. I think this is good to merge

PR #290, opened 2026-02-14
Trim cube at start of postprocessing

thomaswilliamsastro:
- Trim cube at start of postprocessing, rather than at the end
- Add rebin option to trim_cube (True by default)
- Add more robust checks for Stokes axis and various positional axes in trim_cube
- Overwrite should now actually work as intended in trim_cube


thomaswilliamsastro:
Final cube looks a-OK on the edges, so ready for review!
image

PR #293, opened 2026-02-21
Add TestImagingHandler 

e-koch:
Add TestImagingHandler and associated utilities to test different gridding parameters and generate a summary report.

This provides a method to create several image versions and quantify different PSF shape metrics. This is done by selecting a single channel only for lines with no deconvolution. By default the first channel is used, assuming the staged MS parameters pads the velocity range a bit and the first channel will have no signal (for a reliable RMS estimate).

### DIFF
--- a/phangsPipeline/utilsTestImagingPlots.py
+++ b/phangsPipeline/utilsTestImagingPlots.py
@@ -1,0 +1,1807 @@
+"""utilsTestImagingPlots
+
+Utility functions for creating diagnostic plots from test imaging results.
+Creates 1D radial PSF profile galleries, 2D PSF image galleries, and
+2D dirty image galleries for comparing different imaging parameters.
+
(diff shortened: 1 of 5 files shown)


phangsPipeline/utilsTestImagingPlots.py
```diff
@@ -0,0 +1,1807 @@
...
+        Target name for the report title.
+    title : str, optional
+        Custom report title.  If None, auto-generated from target.
+
+    Returns
+    -------
+    str or None
+        Path to the written HTML file.
+    """
+    import base64
```
thomaswilliamsastro: imports normally at the top, I don't see why they shouldn't be in this case

e-koch: I don't mind this if they're only used in a single function. However these are standard packages so agreed they should be at the top

PR #308, opened 2026-03-20
bug in shuffle fixed

akleroy:


akleroy (commented):
Bug in shuffle where nchunk for zipping could go to 0. Fixed, no conflicts. Merging.

PR #312, opened 2026-04-01
Initial noise convergence test implementation and outline noise-only simulated obs for testing

e-koch:
Add additional convergence test based on sampling from the noise distribution.

The main thing this PR adds is `check_noise_convergence`. It's an optional, additional, convergence check that handles cases where a few clean iterations at the noise threshold are included in multiple imaging loops, leading to large fractional changes in the total flux despite being consistent with the noise. We've collectively started hitting chunked imaging cases where a noise-only channel chunk uses all 20 imaging loops and taking way longer to exit than should be needed.

The thinking behind this new test is:
- clean components represent the extremum of the residual image at each iteration. 
- For n_pix finite pixels the extremum is drawn from both tails of the Gaussian, so Var(extremum) ≈ 2·ln(2·n_pix)·(gain·noise)^2.
- the sum S of N_total such extremal draws has Std(S) = sqrt(N_total) · gain · noise · sqrt(2·ln(2·n_pix)).
- we compute the z score = total_flux_iter / Std(S). The criterion passes (null hypothesis) if `z_score < z_threshold`.
-  The threshold is set in `run_imaging` and `loop_imaging` from the image handler with kwarg `convergence_noise_z_threshold`. By default it is None, and this test is not included; passing any finite >0 value enables the noise convergence test, _in addition to_ the existing total flux convergence.

We need the info dict from `fullsummary=True` in tclean and sdintimaging to know the number of iterations actually used, flux per major cycle, etc. The score above defaults to using the last 3 major cycles. There's a future option to correct by scale for multiscale cleaning as this info is also returned in the dictionary.

Trying this on a VLA HI test case for a cube with some low S/N detections, the noise Z-score criterion matches the total flux convergence (the 4 loops is the minimum hardcoded in the PHANGS recipe; otherwise both tests would pass after loop 2):
```
loopnum, deconvolver, niter, cycleniter, threshold, noise, model_flux, frac_delta_flux, noise_z_score
# column 1: Loop number.
# column 2: Deconvolver used in clean.
# column 3: Allocated number of iterations.
# column 4: Cycleniter used to force major cycles.
# column 5: Threshold supplied to clean.
# column 6: Noise level measured in residuals.
# column 7: Integrated model flux.
# column 8: Fractional change in flux from previous loop.
# column 9: Noise z-score.
0, multiscale, 100, 100, 0.02831184606846778Jy/beam, 0.007077961517116945Jy/beam, 1.3617980759704822Jy*chan, inf, 31.32496804221372
1, multiscale, 200, 200, 0.027029895700661062Jy/beam, 0.0067574739251652655Jy/beam, 1.8751552645808185Jy*chan, 0.37697012330149865, 9.468209524700818
2, multiscale, 400, 300, 0.025975869586093447Jy/beam, 0.006493967396523362Jy/beam, 1.8749897100265613Jy*chan, 8.828845130014407e-05, -0.005813716459525169
3, multiscale, 800, 400, 0.025743743418845406Jy/beam, 0.0064359358547113514Jy/beam, 1.8879628674496434Jy*chan, 0.006919055263987687, 0.6851589398047857
```

### DIFF
--- a/phangsPipeline/casaImagingRoutines.py
+++ b/phangsPipeline/casaImagingRoutines.py
@@ -406,20 +406,119 @@ def execute_clean_call(
         if active_kwargs['pblimit'] > active_kwargs['pbmask']:
             active_kwargs['pbmask'] = active_kwargs['pblimit']
 
+    active_kwargs['fullsummary'] = True
+
     if imaging_method == 'tclean':
         # os.mkdir(clean_call.get_param('imagename')+'.image'+'.touch') #<TODO><DEBUG><DL>#
-        casaStuff.tclean(**active_kwargs)
+        tclean_result = casaStuff.tclean(**active_kwargs)
         # os.rmdir(clean_call.get_param('imagename')+'.image'+'.touch') #<TODO><DEBUG><DL>#
     elif imaging_method == 'sdintimaging':
-        casaStuff.sdintimaging(**active_kwargs)
+        tclean_result = casaStuff.sdintimaging(**active_kwargs)
+    else:
+        logger.error('Unexpected imaging method %s' % imaging_method)
+        raise Exception('Unexpected imaging method %s' % imaging_method)
 
     if clean_call.logfile != None:
         casaStuff.casalog.setlogfile(oldlogfile)
 
-    return
+    return tclean_result
 
 
 # endregion
+
+
+def check_noise_convergence(tclean_summary, noise, gain, last_n_cycles=3, z_threshold=2.0,
+                            verbose=False, n_pix=None):
+    """
+    Statistical convergence test: are the recent CLEAN components consistent
+    with noise rather than signal?
+
+    Under H0 (noise only), CLEAN selects the extremum of the residual image at
+    each minor cycle. For n_pix finite pixels the extremum is drawn from both
+    tails of the Gaussian, so Var(extremum) ≈ 2·ln(2·n_pix)·(gain·noise)^2.
+    The sum S of N_total such extremal draws has
+    Std(S) = sqrt(N_total) · gain · noise · sqrt(2·ln(2·n_pix)).
+    When n_pix is None the older formula Std(S) = sqrt(N_total)·gain·noise
+    is used as a backward-compatible fallback.
+
+    Parameters
+    ----------
+    tclean_summary : dict
+        Return value of tclean with fullsummary=True.
+    noise : float
+        RMS noise estimate (Jy/beam), e.g. from noise_for_cube().
+    gain : float
+        CLEAN loop gain (typically 0.1).
+    last_n_cycles : int
+        Number of trailing major cycles from the tclean call to include.
+    z_threshold : float
+        |z| < z_threshold declares components consistent with noise (converged).
+    verbose : bool
+        Print summary info.
+    n_pix : int or None
+        Number of finite pixels in the primary-beam image. When provided,
+        enables the extreme-value variance correction. When None, falls back
+        to the simpler random-sampling formula.
+    Returns
+    -------
+    dict with keys: z_score, N_components, flux_sum, std_expected, converged
+    None if summary is unavailable or malformed.
+    """
+
+    if tclean_summary is None or not isinstance(tclean_summary, dict):
+        return None
+    if 'summaryminor' not in tclean_summary:
+        return None
+
+    summaryminor = tclean_summary['summaryminor']
+    total_flux_increment = 0.0
+    total_iter_done = 0
+
+    for field in summaryminor:
+        for chan in summaryminor[field]:
+            for pol in summaryminor[field][chan]:
+                data = summaryminor[field][chan][pol]
+                try:
+                    iter_done = np.array(data['iterDone'])
+                    model_flux = np.array(data['modelFlux'])
+                    start_flux = np.array(data['startModelFlux'])
+                except (KeyError, TypeError):
+                    continue
+                n_cycles = len(iter_done)
+                if n_cycles == 0:
+                    continue
+                window = min(last_n_cycles, n_cycles)
+                total_flux_increment += float(
+                    np.sum(model_flux[-window:] - start_flux[-window:]))
+                total_iter_done += int(np.sum(iter_done[-window:]))
+
+    if total_iter_done == 0:
+        return {'z_score': 0.0, 'converged': True,
+                'N_components': 0, 'flux_sum': 0.0, 'std_expected': 0.0}
+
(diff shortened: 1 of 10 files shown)


e-koch:
_Local version runs. However, my test case (snapshot VLA in C config) is biased to mean != 0 and isn't well suited as a test. Still working on a suitable noise-only test case._ **This is now corrected as the initial version of the z-score did not account for the number of pixels (draws) in the image in the expected variance.**

phangsPipeline/casaImagingRoutines.py
```diff
@@ -408,19 +408,115 @@ def execute_clean_call(
 
     if imaging_method == 'tclean':
         # os.mkdir(clean_call.get_param('imagename')+'.image'+'.touch') #<TODO><DEBUG><DL>#
-        casaStuff.tclean(**active_kwargs)
+        active_kwargs['fullsummary'] = True
+        tclean_result = casaStuff.tclean(**active_kwargs)
         # os.rmdir(clean_call.get_param('imagename')+'.image'+'.touch') #<TODO><DEBUG><DL>#
     elif imaging_method == 'sdintimaging':
         casaStuff.sdintimaging(**active_kwargs)
+        tclean_result = None
```
e-koch: @thomaswilliamsastro here's the first clear case I've run into where we will need the default sdintimaging. Do we know if the prior issue was corrected upstream? (I suspect so but can't recall what the original issue was)

e-koch: FYI my test case using the casa sdintimaging is running without issue.

e-koch:
My local tests (full imaging case and the new unit tests) are passing

e-koch:
The PR also includes additional tests that we can wrap into a broader testing suite:
- `noise_only_ms_factory.py` produces a very small ALMA 12m Band 6 simulated obs that is purely simulated noise
- `minimal_key_writer.py` sets up mock keys for the simulated noise-only obs (expandable for other test cases, too)
- 2 new test files for the noise convergence: one is unit tests for the `check_noise_convergence` w/o CASA; the other is a full pipeline run through staging and imaging with and without including the new convergence test. These all pass locally but more work is needed to get these and others working in the test suite.

phangsPipeline/casaImagingRoutines.py
e-koch: Patched based on the change within CASA: https://open-bitbucket.nrao.edu/projects/CASA/repos/casa6/diff/casatasks/src/private/task_sdintimaging.py?until=ad32f9697bdb372f975b0c8d11081e0d5c3a9ad6

We will need to revisit this in the future, but for now this works.

thomaswilliamsastro:
Just run through with this. I've pushed the fix for the "no clean components" case to this as well, and it all seems quite a bit faster -- it's running waaaay fewer major cycles. I used `convergence_noise_z_threshold=2`, but I'm not really sure what an appropriate value here is. Might be worth adding in a docstring or something?

I'll have a closer look through the code early next week, but this seems extremely promising!

e-koch:
lower `convergence_noise_z_threshold` is less likely to trigger convergence. I'd say Z=2-3 is a reasonable choice that is unlikely to prematurely exit before convergence is reached. that lines up with the values I'm seeing (above) from some test cases.

That being said, we should absolutely test this to set a reasonable default.

phangsPipeline/casaStuff.py
```diff
@@ -204,3 +204,6 @@ def parse(self, vstr):
 # sdintimaging import
 if (casa_version[0] >= 6):
     from .taskSDIntImaging import sdintimaging
+    # TODO: switch to casatask version of sdintimaging
```
thomaswilliamsastro, line 207: We can *probably* remove this for now...

phangsPipeline/handlerImaging.py
```diff
@@ -190,6 +190,7 @@ def loop_imaging(
                 do_revert_to_singlescale=False,
                 do_export_to_fits=False,
                 convergence_fracflux=0.01,
+                convergence_noise_z_threshold=None,
```
thomaswilliamsastro, line 193: Suggest adding in a bit in the docstring about a recommended value here (I guess this is probably 2 until we do some more detailed testing)

phangsPipeline/handlerImagingChunked.py
```diff
@@ -328,6 +328,7 @@ def run_imaging(
                 do_export_to_fits=True,
                 do_cleanup=True,
                 convergence_fracflux=0.01,
+                convergence_noise_z_threshold=None,
```
thomaswilliamsastro, line 331: Suggest adding in a bit in the docstring about a recommended value here (I guess this is probably 2 until we do some more detailed testing)

phangsPipeline/casaImagingRoutines.py
```diff
@@ -823,6 +929,46 @@ def clean_loop(
...
+            n_pix = None
+            _imagename = working_call.get_param('imagename')
+            pb_image = (_imagename + '.pb') if _imagename else None
+            if pb_image and os.path.isdir(pb_image):
+                try:
+                    myia = au.createCasaTool(casaStuff.iatool)
+                    myia.open(pb_image)
+                    pb_data = myia.getchunk()
+                    myia.close()
+                    n_pix = int(np.sum(np.isfinite(pb_data)))
```
thomaswilliamsastro, line 946: Should this be considering the total pixels within the image, or the total pixels within the mask? I get the feeling it's the former here but my brain is saying it should be the latter

thomaswilliamsastro:
@e-koch Had a look through, there's a couple of comments I had. I think the last one is giving me the most confusion, but maybe I'm being dumb here

I've also done a quick check on the cubes with/without the criteria, and they look identical to me

phangsPipeline/handlerImaging.py
e-koch: added

e-koch:
@thomaswilliamsastro addressed your comments. Should be good to merge now

thomaswilliamsastro (changes requested):

thomaswilliamsastro:
sorry, realised that I hadn't actually finished the review so this comment hadn't appeared. @e-koch it's just some clearup on "which pix" that I don't have clear in my head

phangsPipeline/casaImagingRoutines.py
e-koch: since the mask is a S/N cut based on a similar noise estimate, I think this should default to the whole image to effectively use all the independent "noise" draws. a small number of pixels will make the variance estimates unstable.

thomaswilliamsastro (approved):

PR #322, opened 2026-04-13
Enable scm versioning

e-koch:


e-koch:
@thomaswilliamsastro This enables the scm versioning I had in mind.

We will have to tag a v4.0, though, I was wrong.

phangsPipeline/__init__.py
```diff
@@ -7,8 +7,8 @@
 spectral_cube_enabled = is_spectral_cube_installed()
 
 try:
-    __version__ = version(__name__)
-except PackageNotFoundError:
+    from .version import version as __version__
```
thomaswilliamsastro, line 10: will this work? We don't have a version.py file any more

thomaswilliamsastro: or does it pull in from pyproject

pyproject.toml
```diff
@@ -75,3 +76,7 @@ requires = [
 ]
 
 build-backend = "setuptools.build_meta"
+
+[tool.setuptools_scm]
+# Configure custom options here (version schemes, file writing, etc.)
+version_file = "phangsPipeline/version.py"
```
thomaswilliamsastro, line 82: this file doesn't exist

thomaswilliamsastro (dismissed):

phangsPipeline/__init__.py
e-koch: scm creates the file. It's not meant to be committed.

pyproject.toml
e-koch: same as above: this gets created during the build.

thomaswilliamsastro (approved):

thomaswilliamsastro:
Alright, I just don't understand how this works. I'll merge and then you tag, @e-koch?

Issue #276, opened 2025-12-05
Memory errors processing large data cubes (>20GB)


EOakes:
**Problem**
When processing large ALMA CO data cubes (>20GB, ~4096×4096×500 channels), several functions cause memory errors or system crashes by attempting to load entire cubes into memory at once. 
Running CASA 6.6.1.17 on a system with 250GB RAM. 

**Affected Functions**
`casaMosaicRoutines.generate_weight_file()` - Uses `myia.getchunk()` to load entire cube
`casaMosaicRoutines.mosaic_aligned_data()` - Uses `myia.replacemaskedpixels()`; also `immath()` LEL expressions fail on large image lists
`scConvolution.coverage_collapser()` - `cube.sum` fails on large cubes without `allow_huge_operations = True`
`scConvolution.smooth_cube()` - `cube.convolve_to()` loads entire cube for convolution

Likely non-exhaustive list; this is just what broke my data processing. This may be addressed with a move to SpectralCube + dask; in the meantime they're all patchable with channel-wise processing.

EOakes:
Should be resolved with PR #323

PR #323, opened 2026-04-13
Fix large-cube memory issues with channel-wise processing

EOakes:
Addresses Issue #276 by adding large-cube-safe, channel-wise processing in a few steps that were previously loading too much data into memory. 

Changes: 
- casaMaskingRoutines.py: add channel-by-channel path in noise_for_cube().
- casaMosaicRoutines.py: add channel-by-channel paths in generate_weight_file() and mosaic_aligned_data().
- scConvolution.py: enable huge operations in coverage_collapser(); add channel-by-channel path in smooth_cube().
- handlerDerived.py: add huge cube flag for channel-wise processing.

Why:
These changes prevent memory-related crashes on very large ALMA cubes while preserving existing workflow behavior.

Validation:
Tested on NGC 4945 CO cube in CASA 6.6.1.17 where these steps previously failed due to memory limits.

### DIFF
--- a/phangsPipeline/handlerDerived.py
+++ b/phangsPipeline/handlerDerived.py
@@ -28,6 +28,7 @@ import numpy as np
 
 from . import handlerTemplate
 from . import utilsFilenames
+from .casaCubeRoutines import check_getchunk_putchunk_memory_issue
 from .scConvolution import smooth_cube
 from .scMaskingRoutines import recipe_phangs_strict_mask, recipe_phangs_broad_mask
 from .scMoments import moment_generator
@@ -696,8 +697,10 @@ class DerivedHandler(handlerTemplate.HandlerTemplate):
                 os.system('cp -r ' + indir + input_file + ' ' + outdir + outfile)
 
             else:
-                huge_cube = True
-                logger.debug(f"Setting huge_cube to {huge_cube} for channel-wise processing")
+
+                huge_cube_flag = check_getchunk_putchunk_memory_issue(indir + input_file)
+                if huge_cube_flag:
+                    logger.info(f"Setting huge_cube_flag to {huge_cube_flag} for channel-wise processing.")
 
                 if 'tol' in convolve_kwargs:
                     tol = convolve_kwargs['tol']
@@ -713,7 +716,7 @@ class DerivedHandler(handlerTemplate.HandlerTemplate):
                                 make_coverage_cube=True, coveragefile=outdir + coveragefile,
                                 collapse_coverage=True, coverage2dfile=outdir + coverage2dfile,
                                 overwrite=overwrite, 
-                                huge_cube=huge_cube)
+                                huge_cube=huge_cube_flag)
 
                 if res_type == 'phys':
                     this_distance = self._kh.get_distance_for_target(target)
@@ -728,7 +731,7 @@ class DerivedHandler(handlerTemplate.HandlerTemplate):
                                 make_coverage_cube=True, coveragefile=outdir + coveragefile,
                                 collapse_coverage=True,
                                 overwrite=overwrite, 
-                                huge_cube=huge_cube)
+                                huge_cube=huge_cube_flag)
 
         return ()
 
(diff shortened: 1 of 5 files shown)


thomaswilliamsastro:
I've made some quick edits here, since there were some things that looked duplicated to me and a missing import. You should double-check this all runs on your end. There's one comment I have that might need a bit more work

phangsPipeline/handlerDerived.py
```diff
@@ -696,6 +696,8 @@ def task_convolve(
                 os.system('cp -r ' + indir + input_file + ' ' + outdir + outfile)
 
             else:
+                huge_cube = True
```
thomaswilliamsastro: Can we set this programatically, rather than just forcing it?

thomaswilliamsastro (changes requested):

thomaswilliamsastro:
Also gonna page @e-koch on this

phangsPipeline/handlerDerived.py
e-koch: Agreed though I think it's all good to have `huge_cube=True` be the default kwarg input.

phangsPipeline/scConvolution.py
```diff
@@ -186,10 +251,22 @@ def smooth_cube(
...
+            import os
+            nchan, ny, nx = cube.shape
+            hdr = cube.header.copy()
+            if overwrite and os.path.exists(outfile):
+                os.remove(outfile)
+            fits.PrimaryHDU(np.zeros((nchan, ny, nx), dtype=dtype),
+                            header=hdr).writeto(outfile)
+            with fits.open(outfile, mode='update', memmap=True) as hdul:
+                for i in range(nchan):
+                    hdul[0].data[i] = np.array(cube.filled_data[i], dtype=dtype)
```
e-koch:
```suggestion
                    hdul[0].data[i] = cube.unitless_filled_data[i].astype(dtype)
```
I think avoids a copy/assigning a unit that `np.array` will strip anyways

e-koch:
Looks good @EOakes; thank you! Added a couple small suggestions

thomaswilliamsastro:
I mentioned this to @EOakes on Slack, but I actually ran into a case where this is necessary last week. Can confirm these fixes work on my end, though it would be nice to not force the plane-by-plane if we don't need it, since in my case I/O is quite slow and this slows things down a bit

phangsPipeline/handlerDerived.py
EOakes: Done - flag now set using ccr.check_getchunk_putchunk_memory_issue.

phangsPipeline/handlerDerived.py
```diff
@@ -697,6 +698,10 @@ def task_convolve(
 
             else:
 
+                huge_cube_flag = check_getchunk_putchunk_memory_issue(indir + input_file)
```
thomaswilliamsastro: just rename this to huge cube?

thomaswilliamsastro: Also, ``check_getchunk_putchunk_memory_issue`` relies on CASA, which we don't insist is installed for the derived pipeline. I'd suggest either reading in the fits file we have here and getting the data shape, or just getting the size of the file with ``os.path.getsize`` and if it's bigger than a few GB, flip to huge cube. Don't know if @e-koch has any better thoughts here

thomaswilliamsastro (changes requested):

phangsPipeline/handlerDerived.py
EOakes: Makes sense. Any threshold preference? `check_getchunk_putchunk_memory_issue` uses 2880x2880x393 (~3e9) but I'm not sure where that's from. We could also check the available memory with something like `psutil` but that may be over-engineered.

thomaswilliamsastro: I would maybe stick with those same numbers just for consistency. Loading in a fits file that big should be fine, I'd guess it'd be order 10GB or so

EOakes: Sounds good. Just pushed a commit using `os.path.getsize` - chose this over reading in fits file since it's a little simpler and doesn't have additional dependencies. But can switch to fits implementation if preferred.

thomaswilliamsastro: Seems fine to me, I'm not sure exactly the mapping between image size and no. of bytes but hey, this is about 3GB so seems like the regime where we should be flipping the switch. I'm happy with this, any other worries @e-koch ?

thomaswilliamsastro (approved):

phangsPipeline/handlerDerived.py
e-koch: I think @1054 worked out the threshold where the memory issues were happening, so best to stick with that threshold throughout. Otherwise LGTM!

PR #324, opened 2026-04-14
Add initial CASA tests

thomaswilliamsastro:
This PR adds a bunch of infrastructure for testing code and running end-to-end regressions tests. For now, it's pretty barebones, but should make it easy to add tests going forwards.

Tests are run using pytest, and a tox matrix to test different OS's (and potentially python versions) so we can check if/when things break. Coverage is handled by codecov.io, and code quality by ruff. For now, the ruff doesn't fail on errors since there are a lot, but my plan is to squash these in future PRs and align everything to a more common code practice.

The tests at the moment just check the pipeline imports, CASA imports, and analysisUtils imports. For many of these, it also checks that the arguments we pass exist within the function. So, I see this as a baseline into which we should also include end-to-end regression tests and various tests for functions with known outputs. But it's a lot of code!

- Add pytest suite for testing CASA imports and arguments
- Add pytest suite for testing pipeline imports
- Add pytest suite for testing analysisUtils functions and arguments
- Include coverage
- Run ruff for code checks
- Add tox.ini configuration to run a matrix of tests
- Add GH action for running tests automatically
- Integrate with codecov.io
- Make sure DerivedHandler is only imported if spectral-cube is installed
- Fix changelog entry for dependabot PR
- Tidy GH actions to avoid multiple runs etc.

### DIFF
--- a/tests/casa_config_files/startup.py
+++ b/tests/casa_config_files/startup.py
@@ -1,0 +1,10 @@
+import os
+import sys
+
+envdir = os.environ.get("ENVDIR", os.getcwd())
+
+au_dir = os.path.join(envdir, "analysis_scripts")
+sys.path.append(au_dir)
+
+import analysisUtils as au
+aU = au
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,10 +3,11 @@ name: Build
 on:
   push:
     branches:
+      - master
+    tags:
       - '*'
   pull_request:
-    branches:
-      - master
+  workflow_dispatch:
 
 jobs:
   job:
@@ -14,9 +15,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        # Versions listed at https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
         python-version: [
-          "3.12",
+          '3.12',
         ]
     steps:
       - uses: actions/checkout@v6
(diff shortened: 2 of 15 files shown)


thomaswilliamsastro:
@e-koch ready for review!

tests/casa_config_files/startup.py
```diff
@@ -0,0 +1,10 @@
+import os
+import sys
+
+envdir = os.environ.get("ENVDIR", os.getcwd())
+
+au_dir = os.path.join(envdir, "analysis_scripts")
+sys.path.append(au_dir)
+
+import analysisUtils as au
+import analysisUtils as aU
```
e-koch: do we need both?

Or `aU = au`

e-koch (approved):

tests/casa_config_files/startup.py
thomaswilliamsastro: It's copied from here: https://casaguides.nrao.edu/index.php/Analysis_Utilities (and this is what I've put in my startup.py files for a while). It doesn't look like it should be needed to me, but who ever knows with CASA

e-koch: then we can remove it. It's just a local namespace that we don't use. (Not that it really matters... it's just cleaner)

thomaswilliamsastro: Fixed, I went the aU = au route, just to be safe

PR #343, opened 2026-05-18
Fix crash for different atm types in casaSingleDishALMAWrapper.py

### LINKED ISSUE #311: TP Pipeline bug: No valid MSs found for imaging
It seems that sometimes some files can be named differently to how we'd expect. I've hit:

```
[WARN] [          runALMAPipeline] Did not find uid___A002_X130f861_X4c3d.ms.atmcor.atmtype1_bl
[WARN] [          runALMAPipeline] Did not find uid___A002_X13128dc_X10c80.ms.atmcor.atmtype1_bl
[WARN] [          runALMAPipeline] Did not find uid___A002_X1314bf3_X186e.ms.atmcor.atmtype1_bl
[WARN] [          runALMAPipeline] Did not find uid___A002_X1314bf3_X2ae8.ms.atmcor.atmtype1_bl
[INFO] [          runALMAPipeline] No valid MSs found for imaging.
```

The filenames are actually ``uid___a002_x1314bf3_x2ae8.ms.atmcor.atmtype4_bl`` (4, not 1). Don't know how widespread this is or if there are other variations

thomaswilliamsastro:
Fixes #311.

This PR fixes a crash when TP data uses different atmospheric types. I think this happens in the `hsd_atmcor` call in the TP pipeline but what matters is that dependending on the model, the output filename is different. This just fixes that to search more generally. It'll warn if no files are found for a particular EB, and will raise an Error if multiple files are found for a particular EB.

- Use glob to find baselined files for different atmtypes
- Updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Don't retrigger workflows for dependabot (#332).
 - check-changelog action now uses CHANGELOG.md (#331).
 - Large-cube memory issues with channel-wise processing (#323)
+- Fixed TP crash when different atmospheric correction types are used (#343).
 
 ### Dependencies
 - Bump actions/upload-artifact from 6 to 7 (#313).
--- a/phangsPipeline/casaSingleDishALMAWrapper.py
+++ b/phangsPipeline/casaSingleDishALMAWrapper.py
@@ -293,18 +293,31 @@ def runALMAPipeline(path_galaxy,
     finally:
         h_save()
 
-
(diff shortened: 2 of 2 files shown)


thomaswilliamsastro:
@e-koch this one at least should be quick to review :)

low-sky:
I had also encountered this and was vaguely worried about the glob use but had found no other way to predict / determine what the atm type was going to be.

thomaswilliamsastro:
@low-sky it also seems inconsistent! I had a hard crash with one of the M83 TP obs, but when I was testing this the atmtype it landed on was different. CASA!!!

low-sky:
Yeesh. I can see why the trap for the multiple atmtypes for the same EB becomes important.

low-sky (approved):

Issue #342, opened 2026-05-17
Mismatch between 12m/7m data can lead to weird beams


thomaswilliamsastro:
I thought we'd fixed this, but apparently not. 

It seems like in some cases with data that doesn't overlap properly, combining them can lead to single 7m-only channels at the start/end which means the beam ends up being significantly bigger than it should be. We should really get some checks in place here (may also relate to #248)

thomaswilliamsastro:
Fixed by #344

PR #344, opened 2026-05-18
Fix wrong number of channels in regrid when combining different datasets

### LINKED ISSUE #248: Investigate empty channels at start and end of cube
Several reports of extra NaN-only channels at start and end of cube

thomaswilliamsastro:
Fixes #248 and #342.

This fixes the occasional wrong number of channels when doing the mstransform regrid call in data staging. The first issue, occasional blank channels at the start/end of cube was due to an error in calculating the number of channels to pad by. There is also a second issue where combining datasets from different programmes can lead to different numbers of channels, and hence wonky coverage in the start/end channels. This is now fixed by ensuring all input datasets have the same number of output channels.

N.B. this may have some effect in shifting cubes so the systemic isn't quite in the cube centre. I'm only seeing this change the total number of channels by 1 so not a big deal, but we might want to keep an eye out

- Ensure when combining MS's that they all have the same number of output channels
- Fix wrong number of channels in regrid when combining different datasets
- Update changelog

### DIFF
--- a/phangsPipeline/casaVisRoutines.py
+++ b/phangsPipeline/casaVisRoutines.py
@@ -1040,6 +1040,7 @@ def suggest_extraction_scheme(
     scheme = {}
     chan_width_list = []
     binfactor_list = []
+    total_nchans = []
 
     for this_infile in infile_list:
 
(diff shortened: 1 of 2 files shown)


phangsPipeline/casaVisRoutines.py
```diff
@@ -1601,6 +1655,48 @@ def build_mstransform_call(
...
+
+        # Get the total numbers of channels we should expect *before* rebinning
+        total_nchan_unbinned = total_nchan * binfactor
+
+        # If this doesn't match up, then correct
+        if nchan + add_chans != total_nchan_unbinned:
+            add_chans += total_nchan_unbinned - (nchan + add_chans)
+
+    # Cast back to integer
+    add_chans = int(add_chans)
```
low-sky, line 1695: `ceil` first seems like it might be safer in the cases where we end up slightly off integer because of the velocity <-> frequency non-linearity.  I'm not sure it happens but I always get spooked.

low-sky (approved):
No problems but I haven't run all the corner cases.

PR #347, opened 2026-05-25
Update CASA sdintimaging

thomaswilliamsastro:
This PR syncs our bespoke sdintimaging task to the latest CASA one. There have been improvements here, but the underlying non-convergence is still an issue.

We now will automatically set the dishdia based on beam and frequency, gain can be supplied as a parameter in .clean files, or falls to a default of 1.

Also takes the chance to make the sd prepping a function to reuse code between the imaging handler and the chunked imaging handler.

Note this doesn't have keyHandler changes for inputting gain on a per-obs basis. I think this PR is already big enough

- Update taskSDIntImaging
- General tidy-up of sdintimaging logic
- Updated changelog

### DIFF
--- a/phangsPipeline/handlerImagingChunked.py
+++ b/phangsPipeline/handlerImagingChunked.py
@@ -38,6 +38,7 @@ if casa_enabled:
     from . import utilsFilenames
     from . import utilsLines as lines
     from .clean_call import CleanCall, CleanCallFunctionDecorator
+    from .utilsSingleDish import get_dish_diameter, prepare_sd_image, move_sd_psf
 
     class ImagingChunkedHandler(handlerTemplate.HandlerTemplate):
         """
@@ -723,14 +724,15 @@ if casa_enabled:
 
             if self.imaging_method == 'sdintimaging':
                 # Put in sdintimaging specific parameters to point at the SD image and PSF
-
                 image_name = clean_call.get_param('imagename')
                 clean_call.set_param('usedata', 'sdint')
 
-                sdimage_name = image_name + '.sd.cube.image'
-                sdpsf_name = image_name + '.sd.cube.psf'
-                if not os.path.exists(sdimage_name):
-                    sdimage_name = image_name + '.sd'
+                sdimage_name = image_name + '.sd'
+                sdpsf_name = image_name + '.sd.psf'
+
+                # If we haven't already generated the PSF, leave this blank so
+                # the task will do it
+                if not os.path.exists(sdpsf_name):
                     sdpsf_name = ''
 
                 clean_call.set_param('sdimage', sdimage_name)
@@ -767,11 +769,9 @@ if casa_enabled:
         def task_setup_sdintimaging(
                 self,
                 clean_call=None,
-                target=None,
-                product=None,
                 asvelocity=True,
                 overwrite=False,
-                chunk_num=None
+                chunk_num=None,
         ):
             """
             Regrid existing TP image to the staged MS
@@ -779,8 +779,6 @@ if casa_enabled:
             Parameters
             ----------
             clean_call : CleanCall object
-            target : str
-            product : str
             asvelocity : bool
             overwrite : bool
             chunk_num : int or None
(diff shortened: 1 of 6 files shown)


phangsPipeline/utilsSingleDish.py
```diff
@@ -19,6 +31,163 @@ def get_first_arr_val(arr):
...
+    beam_info = myia.restoringbeam(channel=0, polarization=0)
+    myia.close()
+
+    restoring_beam = beam_info["major"]["value"] * u.Unit(beam_info["major"]["unit"])
+    restoring_beam = restoring_beam.to(u.radian).value
+
+    # Convert through to dish diameter in m
+    lam = 3e8 / freq.to(u.Hz).value
+
+    dishdia = 1.22 * lam / restoring_beam
```
low-sky, line 153: Is 1.22 the right prefactor for ALMA? I know it's not for the GBT, for example. It seems like it should usually be 10.7m (https://help.almascience.org/kb/articles/what-primary-beam-does-casa-use-for-alma-12-m-antennas-and-what-is-the-actual-alma-12-m-antenn)

thomaswilliamsastro: do these change between telescopes? It would be good if this could be pulled out of the header, but I'm not sure it's always recorded. Happy to poke on this to see if it actually gets recorded, or just change the number

thomaswilliamsastro: 1.22 is what CASA seems to use under the hood in feather (https://git.astron.nl/tol/casa/-/blob/master/casa5/code/synthesis/MeasurementEquations/Feather.cc#L247). Since that's what sdintimaging is doing, it's maybe best to keep this consistent (even if wrong?)

low-sky: That was my first question, namely how is it used in the feather.  I was digging through the code but you found it first. Second question is more curiosity, namely what does this give for an ALMA 12m dish in practice?  If it's kicking back 12m with the 1.22, that's definitely not the right thing to go into feather and we may need to be more specific about it.

thomaswilliamsastro: I saw it give 11.6m for my test case. The PSF here just determines the weighting for the Fourier combination in the uv overlap region though, right? So getting it a bit wrong isn't catastrophic, it just ends up being a non-optimal combination. Though I don't think we ever pass the number through in the feathering, so we've **never** actually got this right

phangsPipeline/handlerImagingChunked.py
```diff
@@ -1691,6 +1641,18 @@ def recipe_phangsalma_imaging(
...
             for chunk_to_iter in chunks_to_iter:
 
+                # If we're sdintimaging, make sure we can do this, else fall back to tclean
+                if self.imaging_method == 'sdintimaging':
+                    sd_fits_file = self._kh.get_sd_filename(
+                        target=self.target,
+                        product=self.product,
+                    )
+                    feather_config = self._kh.get_feather_config_for_interf_config(interf_config=self.config)
+                    if not sd_fits_file or not feather_config:
```
e-koch, line 1650: I think this is the default we have for non-chunked imaging. Does it make sense to raise an error instead? Or skip the imaging requested.

phangsPipeline/utilsSingleDish.py
e-koch: But the default is to use a given restoring beam first, I think? The feather docs at least specify each image must have a defined beam: https://casadocs.readthedocs.io/en/stable/api/tt/casatasks.imaging.feather.html#casatasks.imaging.feather

e-koch: No, the sdint helper is setting the dish diameter explicitly:https://open-bitbucket.nrao.edu/projects/CASA/repos/casa6/browse/casatasks/src/private/sdint_helper.py#170

Which I now recall @thomaswilliamsastro mentioning. I don't understand why this needs to be forced vs. relying on the SD beam size defined in the header by default

e-koch (approved):
Looks good. I think we can merge to move sscales imaging along, but I remain skeptical about the requirement for defining the SD weighting based on a dish diameter (which is upstream in the CASA task).

PR #349, opened 2026-06-09
Fix stokes passed as integer string to imsubimage in 4D mosaic path

EOakes:
In the channel-by-channel ndim==4 path of mosaic_aligned_data, imsubimage was being called with stokes='0' (an integer index as a string), but it requires a polarization label like 'I'. Fixed by reading the stokes labels from the coordinate system of myia_sum, which is already open at that point.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Large-cube memory issues with channel-wise processing (#323)
 - Fixed TP crash when different atmospheric correction types are used (#343).
 - Calculation of additional number of channels in regrid mstransform call (#344).
+- Fix stokes passed as integer string to imsubimage in 4D mosaic path (#349).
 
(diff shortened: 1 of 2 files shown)


thomaswilliamsastro:
Seems reasonable to me. I wonder why we haven't found this before? Any ideas @EOakes? I would have thought this should affect any mosaics we're trying to run...

thomaswilliamsastro (approved):

EOakes:
> Seems reasonable to me. I wonder why we haven't found this before? Any ideas @EOakes? I would have thought this should affect any mosaics we're trying to run...

I think it's the fact that a) the cube has to be big enough to trigger the memory-managed path, and b) the input images need to be 3D so adddegaxes fires - in previous testing for #323, I may have overlooked the second condition.

thomaswilliamsastro:
@EOakes OK! It's clear to me that we should probably just keep the 4 axes around through postprocessing so we save a bunch of time adding them back in and ripping them out, so this seems like a good point to start

PR #353, opened 2026-06-12
Keeps Stokes throughout postprocessing

thomaswilliamsastro:
This PR fixes the unnecessary slowdowns that come with re-adding degenerate axes throughout postprocessing.

CASA now typically crashes if the images don't have 4 axes in, which it didn't use to. So we now spend a lot of time ripping them out and putting them back in, which can take a long time for big cubes. Now, we just keep the Stokes axis in throughout, dropping when we export to fits at the end.

This also finally fixes the feather/feather_before_mosaic logic:

- If feather is True and feather_before_mosaic is False, it will feather together individual mosaic tiles, as well as mosaicking the interferometric and singledish data before feathering those together for the final mosaic of mosaics. 
- If feather is True and feather_before mosaic is True, then it will instead mosaic together the feathered individual tiles. Note that if feather_before_mosaic is True, it will still create the full singledish image, but will not use this for the final mosaic.

Fixes:
- Fix bug with has_memory_issue call
- Ensure coordinate axes between interferometric/singledish data match up when staging in postprocessing
- Fix pixperbeam in trim cube to actually respect min_pixperbeam
- Remove dropdegaxes calls, since we don't use them any more
- ccr_dropdeg has now been renamed to ccr_importfits, and modified to either import a fits file or just copy over
- Fix logic with feather_before_mosaic

### DIFF
--- a/phangsPipeline/casaCubeRoutines.py
+++ b/phangsPipeline/casaCubeRoutines.py
@@ -102,20 +102,27 @@ def check_getchunk_putchunk_memory_issue(
         list_to_return.append(cube_mask)
     if return_shape:
         list_to_return.append(cube_shape)
-    return tuple(list_to_return)
+
+    list_to_return = tuple(list_to_return)
+
+    # If we're only returning one thing, remove
+    # the list for forward-proofing
+    if len(list_to_return) == 1:
+        list_to_return = list_to_return[0]
+
+    return list_to_return
 
 #endregion
 
 #region Copying, scaling, etc.
 
-def copy_dropdeg(
+def copy_importfits(
     infile=None,
     outfile=None,
     overwrite=False
     ):
     """
-    Copy using imsubimage to drop degenerate axes. Optionally handle
-    overwriting and importing from FITS.
+    Copy file, handling overwriting and importing from FITS.
     """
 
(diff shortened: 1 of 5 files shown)


phangsPipeline/casaCubeRoutines.py
```diff
@@ -557,15 +542,18 @@ def trim_cube(
...
         bmaj = hdr['restoringbeam']['major']['value']
 
-        pix_per_beam = bmaj*1.0 / pixel_as*1.0
+        pixperbeam = bmaj*1.0 / pixel_as*1.0
+
+        # Calculate the rebinning factor, and rebin if >1
+        rebin_factor = int(np.floor(pixperbeam / min_pixperbeam))
 
-        if pix_per_beam > 6:
+        if pixperbeam > 1:
```
low-sky: It seems like the right move here is to test if rebin_factor > 1?  I haven't moved through but it seems lie pixperbeam could be larger than 1 but the flooring might make it 1?

low-sky (approved):

phangsPipeline/casaCubeRoutines.py
thomaswilliamsastro: fixed

PR #360, opened 2026-06-20
BUGFIX: cleanmask support for sdintimaging

low-sky:
The new sdintimaging code crashes out when a cleanmask is used.  It tries to use the joint cube as a header template for aligning the clean mask. That cube doesn't exist at the time when the alignment code is called.  This bug fix uses the regridded single dish cube as a header template instead.


thomaswilliamsastro (approved):

thomaswilliamsastro:
This seems reasonable to me. Just to check, the .int file also doesn't exist at this point? If that does, I'd suggest using that over the .sd, just for readability

low-sky:
It didn't actually exist at this point but agree that int would be best to template off.

PR #362, opened 2026-06-24
Add more control over multiscale clean scales

### LINKED ISSUE #270: Add clean scales as ratios of the beam size
Instead of giving clean scales only in arcsec:
```
interf_config   	  12m       {'clean_scales_arcsec':[0.5,1,2.5,5.0]}
```
enable a mode in Nx the beam:
```
interf_config   	  12m       {'clean_scales_beam_ratio':[0, 1.0, 2.0, 5.0]}
```

This will be more useful for some of the higher-res CO(2-1) mapping that use different 12m configs to reach the same physical scales across the sample.

thomaswilliamsastro:
Fixes #270 

This PR adds two new options for setting multiscale clean scales, ``clean_scales_beam`` and ``clean_scales_auto``.

The list ``clean_scales_beam`` option allows you to define clean scales as multiples of the beam. If both ``clean_scales_beam`` and ``clean_scales_arcsec`` are defined, then these will be both be used to form a single list of clean scales.

The boolean ``clean_scales_auto`` option, allows for fully automatic setting of the clean scales based on the beam and the LAS of the data. If this is set to True, then will add scales of [0, beam FWHM], then continue multiplying by ``clean_scales_auto_factor`` (default 2) up to ``clean_scales_max_las_fraction`` * the LAS of the data (default 1). Note that setting this will override ``clean_scales_arcsec`` and ``clean_scales_beam``.

- Add ``clean_scales_beam``
- Add ``clean_scales_auto``
- If available, pull beam size from image, or fall back to calculating from visibilities
- Clean scales are now rounded, rather than simply cast to int
- Update docs
- Update CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add badges to README.md (#358).
 - Add support for multiple MSs in singledish pipeline (#365).
 - Added .codecov.yml to configure codecov (#364).
+- Add more control over multiscale clean scales (#362).
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


thomaswilliamsastro:
Moving this back to draft following useful feedback at ADR today

thomaswilliamsastro:
Beam is now pulled from image if available, exposed the important controls for the automatic clean scale setting, and docs updated

@low-sky ready for review!

low-sky (approved):

PR #363, opened 2026-06-25
Add tests for utilsLists, and tidy up

thomaswilliamsastro:
- Formatting pass for utilsLists
- Add fail states to functions
- Remove unused lohi_vecs_to_pairs function
- Add tests for functions
- Update CHANGELOG

### DIFF
--- a/phangsPipeline/utilsLists.py
+++ b/phangsPipeline/utilsLists.py
@@ -1,23 +1,39 @@
-"""
-General purpose utilities.
-"""
-
-
 def select_from_list(
-    master_list, 
-    first=None,
-    last=None,
-    skip=[],
-    only=[],
-    loose=True,
-    ):
+    master_list: list,
+    first: str | None = None,
+    last: str | None = None,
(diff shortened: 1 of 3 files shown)


phangsPipeline/utilsLists.py
e-koch, line 1: @thomaswilliamsastro can you confirm the behaviour is unchanged for these functions, and that's what the tests are based on? I'm struggling to follow the changes in logic and whether they're equivalent

e-koch (commented):
See the comment I added. I think this is all good, but the structure of the logic changed a lot in the updated functions and I don't have a great sense whether it's equivalent and just cleaned up.

phangsPipeline/utilsLists.py
thomaswilliamsastro: Yeah, they produce the same things they used to. It's just all been tidied up and reformatted

e-koch: thanks! will merge

e-koch (approved):

PR #365, opened 2026-07-15
Add support for multiple MS singledish processing

### LINKED ISSUE #346: Looks like there is memory if you keep rerunning singledish pipeline without wiping?
Someone else should verify this but it seems to me that if you rerun the singledish pipeline without wiping the processing directory the brightness of the cube becomes pathological (gets much brighter in Jy/beam). We should probably add a wipe of the temp directories and/or a cleanup to the singledish pipeline.

Would be great if someone can verify this.

thomaswilliamsastro:
This PR adds support for cases where we have multiple singledish MSs for a given target. Essentially, this just adds an extra loop for each MS, but also takes some time to clean up the folder structure a little bit.

This also hopefully fixes #346, since the calibrated/ directory will be cleared out on rerunning, so shouldn't keep applying calibration

- handleSingleDish no longer complains about multiple singledish entries
- Each MS is now copied to a separate folder within the processing root
- When running processing, data is now moved from raw/ to calibrated/, to make things cleaner
- If the calibration has been run before, will clear out the calibrated/ directory before running again

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added more detailed instructions for installing analysisUtils (#340).
 - Included initial documentation (#348).
 - Add badges to README.md (#358).
+- Add support for multiple MSs in singledish pipeline (#365).
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


low-sky (approved):
No issues with this that I see.  Thanks @thomaswilliamsastro

PR #366, opened 2026-07-15
Fix crash with swapped Spectral/Stokes axes for large mosaics

thomaswilliamsastro:
This PR fixes a crash that can occur if the spectral/Stokes axes are swapped when creating a large mosaic.

- Fixed crash if spectral/Stokes axis is swapped when making large mosaics

### DIFF
--- a/phangsPipeline/casaMosaicRoutines.py
+++ b/phangsPipeline/casaMosaicRoutines.py
@@ -1134,6 +1134,8 @@ def mosaic_aligned_data(
     local_outfile = os.path.basename(outfile)
     local_maskfile = os.path.basename(mask_file)
 
+    ndim = len(cube_shape)
+
     if not has_memory_issue:
         # Original approach for smaller cubes
         casaStuff.immath(imagename = local_imlist, mode='evalexpr',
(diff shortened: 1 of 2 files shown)


thomaswilliamsastro:
@EOakes it'd be good to get your eyes on this, I think this is basically the same fix as you put in in some other places not too long ago. For some reason, the pipeline spits stuff out as RA/Dec/Stokes/Freq, rather than our normally assumed RA/Dec/Freq/Stokes

phangsPipeline/casaMosaicRoutines.py
EOakes, line 1: Looks good - this approach also matches what `copy_mask` in `casaCubeRoutines.py` already does. 
Would also be good to correct the hardcoded `blc = [0, 0, ichan, istokes]` around line 1332, which I think faces the same issue. 
Other hardcoded lines (~823, 892, 1118) are iterated over symmetrically, so they'll be mislabeled but won't break things if the axes are swapped.

thomaswilliamsastro: Right, I've edited that now so should be fine as well

EOakes (approved):
Looks good to me now.

PR #371, opened 2026-08-11
Added spectral-cube equivalents for postprocessing

### LINKED ISSUE #369: Move to spectral-cube for postprocessing
Particularly for mosaics, doing the postprocessing in CASA seems deathly slow. I think we'd have a good speed-up just moving from CASA to spectral-cube routines, and this also has the upside of removing some CASA dependency from the pipeline (which seems like generally a good thing)

thomaswilliamsastro:
**TL;DR: Stick `postprocessing_method="spectralcube"` into the `loop_postprocessing` call, enjoy much faster postprocessing.**

Fixes #369.

This PR removes the need for CASA in postprocessing. It uses radio-beam, reproject, spectral-cube, and uvcombine. The bonus of this is that keeping track of axes is now significantly easier, the code is much tidier, and we're generally faster since we don't need to fall back to large cube workarounds. There are some plots below that show regression tests between all of the functions, and I've also checked that the fits files output at the end are comparable in all cases.

Adds 3 main new files, scCubeRoutines, scFeatherRoutines, and scMosaicRoutines. These are mostly equivalent replacements for the CASA equivalents, and have mostly the same arguments. They generally use .fits files, aside from the initial `trim_cube` call.

For now, we keep the CASA versions as the defaults, but you can switch by adding `postprocessing_method="spectralcube"` into the `loop_postprocessing` call. I suggest that at some point we deprecate the CASA routines in favour of these, since they appear to be comparable in my test cases and are significantly simpler to extend and faster.

**Notable function changes:**

- In noise calculations, we now use `scNoiseRoutines.mad_zero_centered`. This cleans up the code, but does remove a bit of functionality.
- In `align_to_target`, we drop the interpolation order for the reprojection. Changing this can cause crashes in the astropy reproject routine, so we do not expose this
- `convert_jytok` and `convert_ktoj` have been merged into a general function `convert_units` where the units can be passed along as a string or astropy.units
- `build_common_header` is now significantly simplified, and does not take overrides for `ra_ctr`, `dec_ctr`, `delta_ra` or `delta_dec`. These aren't actually passed through in the pipeline, so functionally this is the same
- `common_grid_for_mosaic` will reproject to the common grid but extract a minimal sub-cube, to lower disk requirements, as well as put out a text file indicating the (x, y) slice the subcube equates to in the full mosaic
- Mosaicking is done channel-by-channel, to lower memory cost. 
- Any reprojections are done channel-by-channel, to lower memory cost. This should be what spectral-cube does under the hood, but for whatever reason doing it explicitly is much more RAM-efficient

**Function regressions:**

Here are plots for the various functions. Left is the "worst slice" of the cube as defined by maximum absolute sum of the difference between CASA/spectral-cube, and right shows the difference and data, with 1/99th percentile spreads. They look pretty good! With the exception of some edge effects that appear with convolution. Note these plots aren't in order of when they actually happen in the pipeline.

convert_jytok
convolve_to_round_beam
feather_two_cubes
generate_weight_file
linmos_aligned
linmos_commonres
mosaicked
prep_sd_for_feather
primary_beam_correct
trim_cube_rebin
trim_cube
weight_aligned

**Speed differences:**

These are listed as casa/spectral-cube, and were tested for one of the 7m NGC5236 mosaics (_1). Note that these should should be best cases for CASA, and when dealing with big cubes spectral-cube should win out more significantly since we don't have to do costly plane-by-plane calculations

- `trim_cube` (no rebin): 2s/2s
- `primary_beam_correct`: 1.5s/0.5s
- `convolve_to_round_beam`: 102s/2s
- `prep_sd_for_feather`: 4s/14s
- `generate_weight_file`: 2.8s/3.5s
- `feather_two_cubes`: 17.8s/6s
- `trim_cube` (rebin): 4s/0.6s
- `align_to_target`: 0.8s/3.4s
- `convert_jytok`: 0.68s/0.23s
- `export_and_cleanup`: 5.1s/1.32s

For mosaic-related functions, I tested times using 2 of the NGC5236 tiles (_1 and _2)

- `common_res_for_mosaic`: 155.9s/3.75s
- `common_grid_for_mosaic`: 46.36s/119s
- `mosaic_aligned_data`: 22.7s/7.9s

Ran the full NGC5236 7m and 7m+TP mosaics for a "real-world" test case. Overall speed up moving away from CASA: ~3x (3966s/1392s)

Tested the NGC5236 12m+7m+tp case, for a large cube check. Overall speed up moving away from CASA: ~2x (336551s/178217s)

Note that this also leaves a smaller disk footprint: For the 7m run, it's a factor of 3 (25GB vs 9GB). For the 12m+7m+tp, it's a factor 2 (2.4TB vs 1.1TB). Most of this comes from the mosaicking no longer requiring reprojection to a full grid, plus not keeping interchangeable .image and .fits files around at the final export step.

**Branches not explicitly tested:**

I've tested this runs end-to-end with the pipeline defaults. There are some potential edge cases that have no been explicitly tested so we need to keep an eye on them. I think this is the full list:

- apodization in `feather_two_cubes`
- `trim_rind` not tested. Is not used in the default pipeline since an output file is never defined

**Overview of changes:**

*handlerPostprocess*

- Removed trim_coarse_beam_edge_channels, since this is now handled upstream
- Removed unused `recipe_convolve_to_scale`
- `task_rename_sdintimaging` is now spectral-cube postprocessing-aware

*scCubeRoutines*

- Added spectral-cube equivalent for `export_and_cleanup`
- Added spectral-cube equivalent for `convolve_to_round_beam`
- Added spectral-cube equivalent of `align_to_target`
- Added spectral-cube equivalent for `primary_beam_correct`
- Added spectral-cube equivalent for `trim_cube`
- Added spectral-cube equivalent of `trim_rind`
- Added spectral-cube equivalents for `convert_jtok` and `convert_ktojy`: `convert_units`

*scFeatherRoutines*

- Added spectral-cube equivalent for `prep_sd_for_feather`
- Added uvcombine equivalent for `feather_two_cubes`

*scMosaicRoutines*

- Added reproject equivalent for `build_common_header`
- Added spectral-cube equivalent for `common_res_for_mosaic`
- Added spectral-cube equivalent for `common_grid_for_mosaic`
- Added spectral-cube equivalent for `generate_weight_file`, which also includes a new, simplified noise calculation routine
- Added spectral-cube equivalent for `mosaic_aligned_data`

*Others*

- Added radio-beam requirement to pyproject.toml
- Added reproject requirement to pyproject.toml
- Added uvcombine requirement to pyproject.toml
- Update docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added .codecov.yml to configure codecov (#364).
 - Add more control over multiscale clean scales (#362).
 - Add tests for utilsLines, and tidy up formatting (#363).
+- Added spectral-cube equivalent routines for postprocessing (#371).
 
 ### Changed
 
--- a/docs/docs_phangsPipeline/handlers/PostProcessHandler.rst
+++ b/docs/docs_phangsPipeline/handlers/PostProcessHandler.rst
@@ -7,5 +7,12 @@ a single big class (the PostProcessHandler). This needs to be attached
 to a keyHandler to access the target, product, and configuration
 keys.
 
+There are two options for how postprocessing is handled. The default
+uses CASA functions, which is the default but can be slow (especially
+for large cubes). The other option is to use the faster spectral-cube
+routines, which are faster and give comparable results, but are less
+well-tested. To change between these, in the ``loop_postprocess`` call,
+set ``postprocessing_method='casa'`` or ``postprocessing_method='spectralcube'``.
+
 .. autofunction:: phangsPipeline.PostProcessHandler.loop_postprocess
     :noindex:
--- a/phangsPipeline/handlerPostprocess.py
+++ b/phangsPipeline/handlerPostprocess.py
@@ -26,6 +26,33 @@ if casa_enabled:
     logger.debug('casa_enabled = True')
 else:
     logger.debug('casa_enabled = False')
+    
+ALLOWED_POSTPROCESSING_METHODS = [
+    "casa",
+    "spectralcube",
+]
+
+def check_files_exist(
+    f,
+    postprocessing_method="casa",
+):
+    """Check if files exist, potentially modifying by postprocess method
+
+    Args:
+        f (str): filename
+        postprocessing_method (str, optional): Postprocessing method.
+            Should be one of "casa", "spectralcube". Defaults to "casa".
+    """
+
+    if postprocessing_method == "spectralcube":
+        f += ".fits"
+
+    files_exist = True
+    if not os.path.exists(f):
+        logger.warning(f"Missing {f}")
+        files_exist = False
+
+    return files_exist
 
 if casa_enabled:
 
@@ -33,6 +60,9 @@ if casa_enabled:
     from . import casaFeatherRoutines as cfr
     from . import casaMosaicRoutines as cmr
     from . import handlerTemplate
+    from . import scCubeRoutines as scr
+    from . import scFeatherRoutines as sfr
+    from . import scMosaicRoutines as smr
     from . import utilsFilenames
     from . import utilsResolutions
     from .clean_call import CleanCall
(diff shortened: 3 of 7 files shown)


thomaswilliamsastro:
Right, this is ready for review now while I run some last timing tests. Pinging @EOakes and @astrojysun

EOakes:
Ran end-to-end on NGC 4945 12m+7m+TP - Five fields (3200x3600 and 4608x4608, 511 channels) plus the linear mosaic, with `postprocessing_method="spectralcube"` and `imaging_method="sdintimaging"`. 

- Timing is definitely greatly improved: postprocess ran 11h, derived ran 36h, but I didn't do a CASA baseline alongside so no direct runtime comparison. 
- Postprocess & derived products completed cleanly and look good. 
- Two minor commits in PR #382 to address an off-by-one error and file cleanup.

EOakes (approved):

PR #376, opened 2026-08-27
Speedup for sdintimaging

thomaswilliamsastro:
This PR adds significant speedups to sdintimaging in two ways:

- First, and easiest, by making sure that various PSF/PB steps are only carried out the first time. These are slow since they often run major cycles, and only need to be done once to generate cubes, so this has now been fixed.
- Second, because ``feather`` forces a regrid under-the-hood, and this is slow, we now use a smarter uvcombine implementation which doesn't necessarily require regridding (but will if absolutely needed).

Two test cases here, the first is a noise-only 10-channel chunk for NGC5236_1, 7m+TP. This should be the more fair test since you might expect recovering the additional flux that sdintimaging gets would take longer. Timings:

- ``tclean``: 649s
- Old ``sdintimaging``: 1953s
- New ``sdintimaging``: 947s

Second test case here was a signal-filled 10-channel chunk for NGC5236_1, 7m+TP. Timings:

- ``tclean``: 1682s
- Old ``sdintimaging``: 5898s
- New ``sdintimaging``: 2191s

Note these are all run on the same machine. So we're about 1.5x slower than tclean, but 2x faster than the previous sdintimaging implementation, and perhaps a little better in the case with emission. Hooray! I think these are lower bounds as well, as I'd expect for larger cubes the gains should be bigger since these parts often become the bottleneck.

Comparing the signal-filled cubes they're extremely similar. Blue shows 1/50/99th percentiles of the (new implementation) data. The black is the difference, with 1/99th percentiles as the shaded region. The differences here I think are mostly due uvcombine vs feather and the fact that we technically now run fewer major cycles, rather than anything particularly deep. The differences are generally pretty small, and the median diff is locked at 0. The cubes look identical by-eye, structure is the same.

sdint_comparison

Also tagging Remy in here, since I think a good chunk of this would be good to feed back into the actual task

Changes:

- In ``setup_deconvolver``, only run ``makePSF``/``makePB`` if calcpsf=True
- In ``setup_deconvolver``, only run a major cycle to make template residual images if calcres=True
- In ``setup_sdimaging``, only generate SD residual/image if calcres=True
- In ``setup_sdimaging``, only create the SD PSF if calcpsf=True
- In main code, only copy restoring beam to residual image if calcres=True
- In main code, only do the initial feather if calcres=True
- In main code, only feather INT/SD PSF and fitPsfBeam if calcpsf=True
- In main code, only do ``cube_to_taylor_sum`` for the PSF if calcpsf=True
- In main code, only do initial ``cube_to_taylor_sum`` for residuals if calcres=True
- Swap out feather for uvcombine
- Use uvcombine replacements for ``feather_residual`` and ``feather_int_sd``, which live inside the new utilsSDIntImaging.py
- Add uvcombine to pyproject.toml

### DIFF
--- a/phangsPipeline/taskSDIntImaging.py
+++ b/phangsPipeline/taskSDIntImaging.py
@@ -23,6 +23,8 @@ from casatools import table
 from casatasks import imstat
 from casatools import synthesisimager,synthesisutils
 
+from .utilsSDIntImaging import feather_int_sd, feather_residual
+
 try:
     from casampi.MPIEnvironment import MPIEnvironment
     from casampi import MPIInterface
@@ -117,7 +119,12 @@ def setup_imager(imagename, specmode,calcres,calpsf,inparams):
 
     return imagertool
 
-def setup_deconvolver(imagename,specmode,inparams):
+def setup_deconvolver(imagename,
+                      specmode,
+                      calcpsf,
+                      calcres,
+                      inparams,
+                      ):
     """
     Cube or MFS minor cycles. 
     """
@@ -130,20 +137,28 @@ def setup_deconvolver(imagename,specmode,inparams):
     deconvolvertool.initializeNormalizers()
     deconvolvertool.setWeighting()
 
+    ### These three should be unncessary.  Need a 'makeimage' method for csys generation.
+    if calcpsf:
+        deconvolvertool.makePSF()  ## Make this to get a coordinate system
+        #deconvolvertool.makeImage('psf', imagename+'.psf')
+        deconvolvertool.makePB()  ## Make this to turn .weight into .pb maps
 
-    ### These three should be unncessary.  Need a 'makeimage' method for csys generation. 
-    deconvolvertool.makePSF() ## Make this to get a coordinate system
-    #deconvolvertool.makeImage('psf', imagename+'.psf')
-    deconvolvertool.makePB()  ## Make this to turn .weight into .pb maps
-
-        ## Initialize deconvolvers. ( Order is important. This cleans up a leftover tablecache image.... FIX!)
+    ## Initialize deconvolvers. ( Order is important. This cleans up a leftover tablecache image.... FIX!)
     deconvolvertool.initializeDeconvolvers()
     deconvolvertool.initializeIterationControl() # This needs to be run before runMajorCycle
-    deconvolvertool.runMajorCycle(isCleanCycle=False) ## Make this to make template residual images.
+
+    if calcres:
+        deconvolvertool.runMajorCycle(isCleanCycle=False) ## Make this to make template residual images.
  
     return deconvolvertool
 
-def setup_sdimaging(template='',output='', inparms=None, sdparms=None):
+def setup_sdimaging(template='',
+                    output='',
+                    calcpsf=True,
+                    calcres=True,
+                    inparms=None,
+                    sdparms=None,
+                    ):
     """
     Make the SD cube Image and PSF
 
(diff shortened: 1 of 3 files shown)


phangsPipeline/taskSDIntImaging.py
```diff
@@ -165,28 +180,30 @@ def setup_sdimaging(template='',output='', inparms=None, sdparms=None):
...
-        sdintlib.create_sd_psf(sdimage=output+'.residual', sdpsfname=output+'.psf')
-
-    ## Apply the pbmask from the INT image cube, to the SD cubes.
-    #TTB: Create *.mask cube  
-
-    sdintlib.addmask(inpimage=output+'.residual', pbimage=template+'.pb', pblimit=pblimit)
-    sdintlib.addmask(inpimage=output+'.image', pbimage=template+'.pb', pblimit=pblimit)
-
+        
+    if calcres:
```
indebetouw: This isn't exactly "calc res" is it?  Maybe a separate parameter would make more conceptual sense, something like "gridSD"?  
Even better would be to check the coordsys of the SD and IF image, and if they're identical, then don't regrid, and not need a parameter at all, just have it auto-check and not to a completely unnecessary regrid.  That auto-check would also protect against a user putting in a differently-shaped SD image but then saying calcres=False, which would lead to all sorts of wrongness I think.

phangsPipeline/taskSDIntImaging.py
```diff
@@ -165,28 +180,30 @@ def setup_sdimaging(template='',output='', inparms=None, sdparms=None):
...
+        sdintlib.regridimage(imagename=sdimage, template=template+'.residual', outfile=output+'.residual')
+        sdintlib.regridimage(imagename=sdimage, template=template+'.residual', outfile=output+'.image')
+    
+        ## Apply the pbmask from the INT image cube, to the SD cubes.
+        #TTB: Create *.mask cube  
+    
+        sdintlib.addmask(inpimage=output+'.residual', pbimage=template+'.pb', pblimit=pblimit)
+        sdintlib.addmask(inpimage=output+'.image', pbimage=template+'.pb', pblimit=pblimit)
+
+    if calcpsf:
```
indebetouw: In a similar vein to the previous comment, I think its better to

 - leave the checkpsf out of the "if calcpsf" conditional, unless it takes a long time to run (haven't looked at what checkpsf does yet)
 - rather than controlling this with the calcpsf input parameter, have an auto-check that (1) if there's an input parameter sdpsf, it does or doesn't already match the target coordsys, and only regrid if it doesn't. (2) of there's no input parameter, then check if the internal one already exists, and if its coordsys is the target coordsys, and if one or the other of those is not true, then create the internal one.

phangsPipeline/taskSDIntImaging.py
```diff
@@ -588,19 +615,30 @@ def sdintimaging(
...
+            feather_residual(
+                int_cube=int_cube,
+                sd_cube=sd_cube,
+                joint_cube=joint_cube,
+                applypb=applypb,
+                inparm=inpparams,
+                mysdintlib=mysdintlib,
+            )
+        
+        if calcpsf:
```
indebetouw, line 734: Are some checks needed here to make sure that if the user requests calc*=False, that the necessary joint files actually exist? I don't know how things will crash downstream if they don't..

phangsPipeline/taskSDIntImaging.py
```diff
@@ -23,6 +23,8 @@
 from casatasks import imstat
 from casatools import synthesisimager,synthesisutils
 
+from .utilsSDIntImaging import feather_int_sd, feather_residual
```
indebetouw, line 26: is the utils something you made or Urvashi already had?  I haven't checked your feather_residual under the assumption that it was already something Urvashi had, albeit perhaps in a different place?

thomaswilliamsastro: These are the versions of existing SDINT_helper() routines. They're basically the same except for the feather being replaced

phangsPipeline/taskSDIntImaging.py
thomaswilliamsastro: I've put in a bunch throughout which should catch everything

phangsPipeline/taskSDIntImaging.py
thomaswilliamsastro: Also done, needed a bit of a logic rework.

phangsPipeline/taskSDIntImaging.py
thomaswilliamsastro: Done. The regrid actually only does the RA/Dec axes, so I check conformity against those for this conditional. I've also made the checks for the uvcombine regrid a bit more strict given that

phangsPipeline/utilsSDIntImaging.py
```diff
@@ -0,0 +1,310 @@
...
+            logger.info("Reprojecting the SD cube spectrally to match INT cube")
+            sdcube_sc = sdcube_sc.spectral_interpolate(intcube_sc.spectral_axis)
+
+        # If we don't have matching RA/Dec, then we'll need to reproject
+        do_spatial_reproject = False
+        if not intcube_sc.shape[1:] == sdcube_sc.shape[1:]:
+            do_spatial_reproject = True
+        else:
+            wcs_conforms = check_wcs_conforms(
+                intcube_sc.wcs.celestial.to_header(),
```
low-sky, line 154: I have only moderate concerns that this WCS checking won't work in cases where it's using the CD matrix representation. If it's all CASA, we should be okay, but if the SD is from somewhere else.

Maybe checking `intcube_sc.wcs.celestial.wcs.compare(sdcube_sc.wcs.celestial.wcs)`?

low-sky: Never mind.  I see it below.

low-sky (approved):

PR #382, opened 2026-08-31
Two small fixes in spectral-cube postprocessing (trim_cube and mosaic files)

EOakes:
1. `trim_cube` off-by-one error drops the final slice in `scCubeRoutines.py` (inclusive vs. exclusive stop). This runs once for prep and once for cleanup, so compounds in the postprocessing stage. 

2. Mosaic text files aren't renamed under the stdintimaging branch (`task_rename_sdintimaging` in 
`handlerPostprocess.py`) to include "tp" (e.g., "12m+7m" -> "12m+7m+tp") and are stranded.  This matters if the mosaic stage is rerun, which would otherwise fail to find the files.

### DIFF
--- a/phangsPipeline/handlerPostprocess.py
+++ b/phangsPipeline/handlerPostprocess.py
@@ -1113,6 +1113,23 @@ if casa_enabled:
                     os.system('rm -rf %s' % new_file_name)
                     os.system(command)
 
+                # Rename the image_slice mosaic files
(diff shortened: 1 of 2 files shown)


thomaswilliamsastro (approved):
Thanks! I didn't explicitly test sdintimaging, and I always mess up on array slicing indices

PR #384, opened 2026-08-31
Fix a wrong key name in the SD pipeline

astrojysun:
In [`handlerSingleDish.py`](https://github.com/astrojysun/phangs_imaging_scripts/blob/js_dev/phangsPipeline/handlerSingleDish.py) the corresponding key is named `max_chanwidth_kms`. We did a ["translation" of parameter names](https://github.com/astrojysun/phangs_imaging_scripts/blob/js_dev/phangsPipeline/handlerSingleDish.py#L289) in the legacy pipeline call but not for the new pipeline, which causes a key error.


e-koch (approved):
Good catch!

thomaswilliamsastro (approved):

PR #385, opened 2026-09-01
Improve spectral-cube convolution

thomaswilliamsastro:
This PR aims to improve and consolidate cube convolution. The main change is we now have a single function for convolution, which operates channel-by-channel in all cases. This simplifies the code and reduces RAM usage, without any noticeable change in speed. It can either use the spectral-cube convolution (which is a wrapper around astropy's convolve/convolve_fft) or a new convolve_uv function that applies a transfer function directly in the uv plane. This means we do not have to pad out the convolution kernel, which leads to a ~10% smaller beam (linear) after all the convolutions we do -- for free!

This convolve method is cleanly exposed in the `loop_postprocess` call, where you can specify `convolve_method=convolve|convolve_fft|convolve_uv`. The same is true in `loop_derived_products`. For consistency, the default is `convolve_fft`.

Note that because we now write out channel-by-channel within the convolution step, an outfile needs to be specified in `scConvolution.smooth_cube`. This is always done in the pipeline, so is not an issue.

- Add `convolve_uv` method to convolution
- Remove `ftplane_convolution.py`
- Exposed convolution keyword for both postprocessing and derived handlers
- Removed `huge_cube` from `handerDerived`
- `scConvolution.smooth_cube` now calls `scCubeRoutines.convolve_cube`
- Make sure we remove references to multibeam in the header during convolution, since we always convolve to a single beam
- `scCubeRoutines.convolve_to_round_beam` now uses radio-beam common beam calculation for varying resolution cubes
- `scMosaicRoutines.common_res_for_mosaic` now uses radio-beam common beam calculation
- Deprecate pixel padding in `scMosaicRoutines.common_res_for_mosaic`
- Fixed missing import for `channel_width` if spectral convolution is requested in `smooth_cube`
- Fixed bug where convolving channel by channel would lead to incorrect values due to not scaling Jy/beam properly. Note this is a bug in spectral-cube, and should be fixed shortly
- Add convolve_uv to pyproject.toml
- Add PyPI version of uv-combine to pyproject.toml
- Pin spectral-cube version, until new version fixes bugs


thomaswilliamsastro:
Run is finished on the M83 12m data. All looks good! Beam is ~4% smaller doing this, which is a small win. New is on the left, old (CASA) on the right. There are some differences here but I think it's mainly due to feather->uvcombine

Screenshot 2026-09-04 at 09 42 55

phangsPipeline/handlerDerived.py
```diff
@@ -725,10 +747,6 @@ def task_convolve(
 
             else:
 
-                huge_cube = os.path.getsize(indir + input_file) >= 2880*2880*393
```
e-koch, line 728: This is because we default to per-channel convolution and avoid having the whole cube in memory?

phangsPipeline/scCubeRoutines.py
```diff
@@ -441,8 +444,7 @@ def convolve_to_round_beam(
     # This is a little different depending on whether the cube is a VaryingResolutionSpectralCube
     # or not
     if isinstance(cube, VaryingResolutionSpectralCube):
-        beam_idx = np.argmax([b.major.to(u.arcsec).value for b in cube.beams])
-        beam = cube.beams[beam_idx]
+        beam = cube.beams.common_beam()
```
e-koch: Is this where we enforce circularizing the beam? The general common beam will tend to be elliptical, in which case sticking to the max major axis is still the minimal circularized beam size.

This does also, occasionally, fail to converge b/c it's an approximate solution. If we do keep it, here's the function that's controlling the solver: https://radio-beam.readthedocs.io/en/latest/api/radio_beam.commonbeam.common_manybeams_mve.html#radio_beam.commonbeam.common_manybeams_mve

`tolerance` and `epsilon` are the kwargs that may need to be exposed.

phangsPipeline/scCubeRoutines.py
```diff
@@ -466,53 +473,165 @@ def convolve_to_round_beam(
     # Build beam, do the convolution, write out
     target_beam = Beam(major=target_bmaj, minor=target_bmaj, pa=0 * u.deg)
 
+    # Sometimes, this can cause some small errors so we add a (very small)
+    # epsilon if the target_beam cannot be deconvolved. This is by definition
+    # not an issue for convolution methods where we pad out the kernel
+    if convolve_fn in ["convolve_uv"]:
+        epsilon = 5e-4
```
e-koch: agreed with this approach! the more general alternative would be epsilon as a fracton of the pixel scale (say 10%).

e-koch: you know, for all the VLBI data we image with the phangs pipeline :)

phangsPipeline/scCubeRoutines.py
```diff
@@ -466,53 +473,165 @@ def convolve_to_round_beam(
...
+                        target_beam=target_beam,
+                        **kwargs,
+                    )
+                elif convolve_fn == "convolve":
+                    cube = cube.convolve_to(
+                        target_beam,
+                        convolve=convolve,
+                        **kwargs,
+                    )
+                elif convolve_fn == "convolve_fft":
```
e-koch: convolve and convolve_fft have the same function call so could be combined together?

phangsPipeline/scCubeRoutines.py
```diff
@@ -466,53 +473,165 @@ def convolve_to_round_beam(
...
+                    cube = cube.convolve_to(
+                        target_beam,
+                        convolve=convolve_fft,
+                        **kwargs,
+                    )
+                else:
+                    raise ValueError(f"Unknown convolution function {convolve_fn}")
+
+                cube_conv = cube.unitless_filled_data[:]
```
e-koch, line 582: Just to remind us these cubes are "2D" and so the read into memory this line triggers is OK.

```suggestion
                # the "cube" is 2D here. This should not require a large memory to read in.
                cube_conv = cube.unitless_filled_data[:]
```

phangsPipeline/scCubeRoutines.py
```diff
@@ -466,53 +473,165 @@ def convolve_to_round_beam(
...
+                        beam_ratio_factor = 1.0
+
+                    # Do the actual convolution
+                    if convolve_fn == "convolve_uv":
+                        chan_slice = convolve_uv(
+                            image=chan_slice,
+                            target_beam=target_beam,
+                            **kwargs,
+                        )
+                    elif convolve_fn == "convolve":
```
e-koch: same comment as above on whether this can be an elif convolve or convolve_fft

phangsPipeline/scMosaicRoutines.py
```diff
@@ -183,14 +178,23 @@ def common_res_for_mosaic(
...
-            bmaj = cube.beam.major
-
-            bmaj_list.append(bmaj.to(u.arcsec).value)
+            beam_list.append(cube.beam)
             pix_list.append(pixel_as.to(u.arcsec).value)
 
-        max_bmaj = np.max(bmaj_list)
+        # Calculate a common beam
+        beam_list = Beams(beams=beam_list)
+        common_beam = beam_list.common_beam()
```
e-koch: same comment on whether we enforce a circular beam here or not. I think we should.

In which case this should just be a beam_list.max() for the smallest common circular beam

pyproject.toml
```diff
@@ -36,15 +36,16 @@ classifiers = [
...
     "matplotlib >= 3.10.9",
     "numpy >= 2.3.5",
     "protobuf == 3.20",
     "pytz >= 2025.2",
     "radio-beam >= 0.3.9",
     "reproject >= 0.21.0",
     "scipy >= 1.16.3",
-    "spectral-cube >= 0.6.7",
-    "uvcombine@git+https://github.com/radio-astro-tools/uvcombine.git",
+    "spectral-cube == 0.6.7",
```
e-koch, line 47: we should really keep >=. I'll release an version in the coming weeks and we're not so careful on that end that pinning to a single version as "stable" should be best practice.

e-koch (changes requested):

phangsPipeline/handlerDerived.py
thomaswilliamsastro: yep, exactly. We're just always doing channel-by-channel so this isn't a thing any more

phangsPipeline/scCubeRoutines.py
thomaswilliamsastro: Right, this does that below to get circular (l474) and then checks we can deconvolve (l479:485)

phangsPipeline/scCubeRoutines.py
thomaswilliamsastro: Done

phangsPipeline/scCubeRoutines.py
thomaswilliamsastro: Done

phangsPipeline/scCubeRoutines.py
thomaswilliamsastro: Done

phangsPipeline/scCubeRoutines.py
thomaswilliamsastro: Done

phangsPipeline/scMosaicRoutines.py
thomaswilliamsastro: It does this (l212), and in practice the common_beam() returned here always seems to just be the largest beam

pyproject.toml
thomaswilliamsastro: This is only pinned for now because of the Jy/beam scaling w/cube slices (l594 in scCubeRoutines has the comment). Once the fix is out, we should remove our fix there, bump this and move back to >=

e-koch: Ok

phangsPipeline/scMosaicRoutines.py
e-koch: right, but the max is more robust against failures for the common beam calcs (the beam deconvolution errors you noted in the radio-beam issue). It's also a lot faster.

b/c we're circularizing the beam, the smallest _circular_ beam has to be the max(bmaj). The smallest common beam is only an advantage when it's allowed to be elliptical.

thomaswilliamsastro: Aight, that's done now

thomaswilliamsastro: Wait, maybe it's not. For some reason beam_list.max() seems to return a list...

e-koch: Hmm. Maybe `beam_list.major.max()`?

thomaswilliamsastro: Ah, it's .largest_beam()

e-koch: probably, but it may  not always be with a highly elliptical beam. all the `Beam.max()` functions are in terms of the beam area; we want the largest major axis.

thomaswilliamsastro: Now using Beams.major.max()

phangsPipeline/scCubeRoutines.py
e-koch: I suggest we switch to `Beams.major.max()` here too, since we're circularizing below

thomaswilliamsastro: Done!

e-koch (approved):

PR #389, opened 2026-09-02
Cleanup and tests for scMoments.py

thomaswilliamsastro:
This PR does a cleanup pass of the code for scMoments and adds tests

- Docstrings tidied up, properly typed, added arguments
- Removed unused `_nicestr` function
- `_func_and_kwargs_for_moment` is now tidied up and more easily extensible
- Removed unused `moment_tag_known` function
- Be more specific about errors in `moment_generator`
- Removed unused `context` and `assignkunits` args from the `moment_generator`
- Remove K conversion in `moment_generator`
- Removed deprecated `scTemp.py`
- Update CHANGELOG.md


phangsPipeline/scMoments.py
```diff
@@ -10,159 +9,156 @@
...
     else:
-        logging.error('Unrecognized input type for cubein')
-        raise NotImplementedError
+        raise TypeError("cubein should be a string or SpectralCube object.")
 
     cube.allow_huge_operations = True
-        
+
     # Force Kelvin. We will be unit agnostic later.
     cube = cube.to(u.K)
```
e-koch: Can we remove this now? We should have handled the K conversion in an earlier step and it keeps these functions more general in case a Jy/beam km/s like moment output is what someone wants out

tests/test_scMoments.py
```diff
@@ -0,0 +1,204 @@
...
+    wcs.wcs.crpix = [0.0, 0.0, 0.0]  # Reference pixel
+
+    # Construct the SpectralCube
+    cube = SpectralCube(data=data,
+                        wcs=wcs,
+                        )
+    cube_file = os.path.join(tmp_path.name, "scMoments_test_cube.fits")
+    cube.write(cube_file, overwrite=True)
+
+    mask = SpectralCube(data=np.ones_like(data),
```
e-koch, line 39: The test should probably have a mask that has some zeros as the typical use case

tests/test_scMoments.py
```diff
@@ -0,0 +1,204 @@
...
+                        wcs=wcs,)
+    mask_file = os.path.join(tmp_path.name, "scMoments_test_mask.fits")
+    mask.write(mask_file, overwrite=True)
+
+    noise = SpectralCube(data=np.random.normal(loc=0, scale=0.3, size=data.shape),
+                         wcs=wcs,
+                         )
+    noise_file = os.path.join(tmp_path.name, "scMoments_test_noise.fits")
+    noise.write(noise_file, overwrite=True)
```
e-koch, line 49: Do these temp files get cleaned up?

e-koch (changes requested):
Couple minor things to check

tests/test_scMoments.py
thomaswilliamsastro: as far as I can tell, yep. Running locally the tmp_path gets deleted once the code is finished running

phangsPipeline/scMoments.py
thomaswilliamsastro: Done

tests/test_scMoments.py
thomaswilliamsastro: Done

e-koch (approved):

PR #390, opened 2026-09-03
Cleanup and tests for utilsResolutions.py

thomaswilliamsastro:
This PR does a cleanup pass of the code for utilsResolutions and adds tests

- Docstrings tidied up, properly typed, added arguments
- Removed many now unused regex patterns
- `is_angular_resolution`,`is_physical_resolution`, `is_distance` have been significantly simplified to use one underlying function, `is_astropy_unit_equivalent`, which makes use of astropy.units to significantly tidy up the code
- Exposed `sig_figs` in `get_tag_for_angular_resolution`, `get_tag_for_physical_resolution`, and `get_tag_for_res`
- Calculation of angular resolution in `get_angular_resolution_from_physical_resolution` is tidied up to use astropy.units
- Update CHANGELOG.md

### DIFF
--- a/phangsPipeline/utilsResolutions.py
+++ b/phangsPipeline/utilsResolutions.py
@@ -1,187 +1,293 @@
-"""Utilities for angular and physical resolutions.
-"""
-
 import re
-import numpy as np
 
-regex_psep = re.compile(r'([0-9eE.+-]+)(p)([0-9eE.+-]+)(.*)', re.IGNORECASE)
-regex_nounit = re.compile(r'^([0-9eE.+-]+)$', re.IGNORECASE)
-regex_arcsec = re.compile(r'^([0-9eE.+-]+)[ \t]*(arcsec|\")$', re.IGNORECASE)
-regex_arcmin = re.compile(r'^([0-9eE.+-]+)[ \t]*(arcmin|\')$', re.IGNORECASE)
-regex_degree = re.compile(r'^([0-9eE.+-]+)[ \t]*(degree|deg)$', re.IGNORECASE)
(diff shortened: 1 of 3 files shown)


phangsPipeline/utilsResolutions.py
```diff
@@ -1,187 +1,293 @@
...
-            raise Exception('Need a distance for calculating angular resolution from the given physical resolution of '+str(res))
-        return get_angular_resolution_from_physical_resolution(res, distance)
+    is_dist, dist_mpc = is_distance(
+        distance,
+        return_value=True,
+    )
+    if not is_dist:
+        raise ValueError(f"Input distance string {distance} is not a valid distance")
 
+    ang_res_arcsec = (phys_res_pc / dist_mpc / 1e6 * u.rad).to(u.arcsec).value
```
e-koch, line 256: Suggest we remove hardcoded unit conversions altogether. Do `phys_res_pc` and `dist_mpc` retain their units to this point?

```suggestion
    ang_res_arcsec = ((phys_res_pc / dist_mpc).to(u.one).value * u.rad).to(u.arcsec).value
```

e-koch (commented):
Significantly simplified! I caught one case where we're still using a hardcoded unit conversion

phangsPipeline/utilsResolutions.py
thomaswilliamsastro: They don't, and since this relies on other functions to parse out the value I think we want to just keep this as is. Agree it's a little clunky

e-koch: then yeah, let's keep as is. Otherwise it's still clunky reassigning the units back

e-koch (approved):

PR #391, opened 2026-09-08
Bugfixing and testing sdintimaging

low-sky:


pyproject.toml
```diff
@@ -122,3 +122,5 @@ build-backend = "setuptools.build_meta"
 [tool.setuptools_scm]
 # Configure custom options here (version schemes, file writing, etc.)
 version_file = "phangsPipeline/version.py"
+tag.strict = true
```
thomaswilliamsastro, line 125: What does this do?

thomaswilliamsastro (approved):

pyproject.toml
thomaswilliamsastro: Not seen builds fail because of tags, but interesting to know

PR #393, opened 2026-09-11
Calculate LAS better in auto-multiscales

thomaswilliamsastro:
This PR adds a new function to `casaVisRoutines`, that calculates the LAS for a dataset in a more robust way. For each observation ID (since we're concatenating a bunch of MS's) it'll calculate the LAS using the working definition in the ALMA handbook (though the parameters are exposed to theoretically tune this). It'll then take the maximum LAS from all of these individual observations and report this. This is necessary to not drown out the 7m baselines when we combine with 12m data.

Testing on one 12m+7m MS:

- `estimate_mrs`: LAS=31arcsec
- `au.estimateMRS`: LAS=9arcsec

This new function gives a result consistent with the just-7m MS, which is what should happen.

The default values for `clean_scales_auto_factor` and `clean_scales_max_las_fraction` have also been updated to roll in some external wisdom. Per the CASA tclean docs, they recommend using a factor of 3-5 for additional clean scales, so we now default to 3. The max LAS fraction now defaults to 0.5, as adding scales up to the full LAS can be dangerous if the baselines are noisy or sparsely sampled. This should be more conservative. Of course, this is still totally configurable, but hopefully represents some best practices. I expect we'll tune this further in the future.

- Add `estimate_mrs` to `casaVisRoutines`
- `clean_call` now uses `estimate_mrs` rather than the analysisUtils function
- Update the defaults for `clean_scales_auto_factor` and `clean_scales_max_las_fraction`
- Update CHANGELOG.md


phangsPipeline/casaVisRoutines.py
```diff
@@ -2315,3 +2317,107 @@ def noise_spectrum(
...
+    Returns:
+        dict: Dictionary containing the representative frequency, baseline for MRS,
+            and the MRS in arcseconds.
+    """
+
+    tb = casaStuff.tbtool()
+
+    # Obtain representative frequency from the average of the frequencies.
+    tb.open(vis + "/SPECTRAL_WINDOW")
+    frequencies = np.asarray(tb.getcell("CHAN_FREQ"), dtype=float)
```
e-koch: Isn't there a column with the frequency at the central channel?

phangsPipeline/casaVisRoutines.py
```diff
@@ -2315,3 +2317,107 @@ def noise_spectrum(
...
+    frequencies = frequencies[np.isfinite(frequencies) & (frequencies > 0)]
+    rep_freq = np.median(frequencies) * u.Hz
+    tb.close()
+
+    tb.open(vis)
+
+    # Pull out information and decide valid antenna pairs.
+    uvw = np.asarray(tb.getcol("UVW"))
```
e-koch: does this read in for every visibility in the table? That could start to get large, at least in some of our use cases (LGLBS M31 ABCD combined data is ~120 milliion rows, even after frequency and time averaging).

If there's a way to make sure we're only reading the unique UVW for a single channel representative channel, that seems like it'll be safer. Same goes for the flags, though those are bools.

phangsPipeline/casaVisRoutines.py
```diff
@@ -2315,3 +2317,107 @@ def noise_spectrum(
...
+        uvw = uvw.T
+
+    uv_distance_m = np.hypot(uvw[0], uvw[1])
+
+    # FLAG normally has dimensions (ncorr, nchan, nrow).
+    # Keep a row if at least one correlation/channel is unflagged.
+    flag_axes = tuple(range(flags.ndim - 1))
+    completely_flagged = np.all(flags, axis=flag_axes)
+
+    valid = (
```
e-koch: this makes me nervous for large data sets. I suggest moving the selection to a relevant sample selection (single channel for representative SPW) to read in only a sub-set of the rows.

e-koch: or even loop by SPW, selecting a central channel for each.

e-koch (commented):
Updates look good. I have some questions about splitting up the reads from the visibility table to guard against memory issues.

phangsPipeline/casaVisRoutines.py
thomaswilliamsastro: Done

phangsPipeline/casaVisRoutines.py
thomaswilliamsastro: Changed this now. It's not a representative channel, since each row is for an antenna pair. Instead, the loop has been changed so we downselect on OBSERVATION_ID first and then (optionally, but on by default) by the first FIELD_ID. For highlighting how this cuts down rows (for a 12m+7m dataset):

```
Initial nrows: 1813095
Downselecting to OBSERVATION_ID: 14976/11648 (7m), 425657/387450/445324/283587/199525
Downselecting to first FIELD_ID: 288/224 (7m), 2838/2583/2970/1892/1560 (12m)
```

So *significantly* cutting down how much we're reading. To guard against loading in loads of the table at once, it also now loops rows in chunks (of 100,000 by default)

phangsPipeline/casaVisRoutines.py
thomaswilliamsastro: See above

phangsPipeline/casaVisRoutines.py
thomaswilliamsastro: Using the first field can cause a bit of a change in the calculated LAS (for a `.split.cal` file I saw it change by ~10%, for a concat'd MS I saw it change by less than 1%), but I think the I/O gains here are probably worth it. Since this is just for setting the MS clean scales I don't think we need to be super exacting

e-koch: Looks good. And there's an implicit assumption already built in that the field-to-field uv coverage needs to be small for the mosiacking w/ tclean to work in the first place.

e-koch (approved):

PR #396, opened 2026-09-15
Improved weights for linear mosaicking

### LINKED ISSUE #372: [BUG] Noise generation in postprocessing should be measured from non-PB corrected image
Currently, we measure the noise for mosaics from the pbcorr files. This is fine if all input mosaic tiles are the same shape, but will cause biases if they're not, due to the noise being non-uniform once primary beam correction has taken place

### LINKED ISSUE #373: [FEAT] Noise generation should be done on _aligned images
Currently, we measure noise on images before they go through a second round of convolution to a common beam for all mosaic tiles. This will cause issues if the common beam is significantly different to the individual image beam. Suggested pathway

-> Regrid PB to common grid
-> Convolve/regrid all mosaic tiles to common grid
-> Perform noise generation using convolved/regridded mosaic tiles and regridded PB

thomaswilliamsastro:
Fixes #372
Fixes #373

This PR improves the weight calculations for interferometric data and tidies up the logic a bit. We now calculate the weights in the mosaicking step, so they're not generated if we're not making a mosaic. The weights are based on the *_aligned files (i.e. after convolution and regridding), and after undoing primary beam corrections if applicable.

N.B. the undoing pbcorr only works in the spectralcube postprocessing branch. I think this is an admission that we are starting to diverge and at some point should fully deprecate the casa routines. Because I'm lazy and don't want to maintain 2 codebases.

- Generate weights in mosaicking steps
- Generate weights on linmos_aligned and sd_aligned images
- Noise calculation will undo primary beam correction before calculating noise, if applicable
- Update CHANGELOG.md


thomaswilliamsastro:
Note that in practice, this doesn't seem to be super important (it's basically small details). Which is good, but we'd be noticing this was nonsense earlier in the mosaics if it wasn't the case

Weight map. Old on left, new on right. It's same percentage scaling, the numbers are ~20% different but all in the same way

Screenshot 2026-09-15 at 15 41 48

And just to show a slice of cube (easier to see the steps between the tiles in a noise channel)

Screenshot 2026-09-15 at 15 43 33

The fluxes are the same to like the 5th decimal place

e-koch (approved):

e-koch:
On starting to diverge from the sc support: does the original method still work, just without the un-pbcorr step?

thomaswilliamsastro:
@e-koch just checked, all works on the CASA branch too

PR #397, opened 2026-09-16
Use weight file from SD pipeline

thomaswilliamsastro:
This PR brings in the weight files that the SD pipeline produces, to improve linear mosaicking.

Left here is what we had previously, middle is using the weight image as is, right is using weight^2. It's a lot better, and I suspect the last part will be improved with improved baselining

Screenshot 2026-09-16 at 13 30 56

- Edit the weight file in the SD pipeline to square what the pipeline spits out, so the weight works as expected
- Pull in the SD weight file, if it exists
- Stage SD weight file like the SD file
- Don't generate weight file if SD weight file already exists
- Updated CHANGELOG.md


thomaswilliamsastro:
@low-sky ready for review!

PR #398, opened 2026-09-18
Try to appropriately change the spectral units for velocity->freq

e-koch:
@indebetouw -- this _should_ fix the SD spectal unit mismatch, assuming well-behaved headers and such.

Trying to automatically pull the correct velocity convention is an upstream problem to handle in spectral-cube, but this is likely to cover all of our current use cases.


thomaswilliamsastro:
Looks reasonable to me, but I'll wait for Remy's testing to confirm this works

